Issue #27, opened 2018-02-03
Lock down / specify versions


jonyardley:
There are a few pre-requisites which currently don't specify version numbers. Let's at least specify them or modify the instructions to lock down versions. We could also include a `.node-version` file etc...

jonyardley:
I think this can be closed now?

robbiemccorkell:
I think so yea

PR #29, opened 2018-02-11
Update depenencies 

robbiemccorkell:
Update dependancies to their latest versions.

Couldn't update flow all the way, because React Native currently expects 0.63.

## Pre-flight check-list

Before raising a pull request

* ~~Documentation~~
* ~~Unit tests~~
* [x] Build passing

## Pre-merge check-list

* [x] Code review (At least one :thumbsup:)
* ~~Tester approved~~


markholland:
Do we want to add .node-version here as well? #27 
I haven't added one yet as I've seen there isn't actually a standard for that file yet and is still under [discussion](https://github.com/nodejs/version-management/issues/13)

robbiemccorkell:
Good point. I pushed one up with a version matching that in Circle. We'll see about format, can always change it later.

PR #32, opened 2018-02-11
[WIP] Integrate PRs with Appetize

robbiemccorkell:
Looking at integrating branch builds with [https://appetize.io/](Appetize) so we can try the features built on a branch without pulling code locally. We might want to run scripted journeys against this system in the near future too.

Crucially once deployed to Appetize, we need to comment back into a PR or set a build status with a link for easy PR review and testing.

## Pre-flight check-list

Before raising a pull request

* [ ] Documentation
* [ ] Unit tests

## Pre-merge check-list

* [ ] Tester approved


robbiemccorkell:
I've _almost_ got this working and reporting back with a build status and link. Parked it briefly to take a break and build some UI, which I think would be more helpful right now because people can start building on top of UI scaffolding.

markholland:
I think this is great and just want to ask if we think this will be suficient for merging approval/testing? Or should I look further at adding deploys to hockeyapp by suffixing the branch or sha to the binary ID so that it's plausible for people to hunt for the build in hockeyapp or have them sent to a Slack channel. Or should we have master be our 'qa' branch and then have release branches? Just trying to picture what the flow would look like for an individual feature.

robbiemccorkell:
Well this will be useful, but a physical device will always be good for testing. Do you think we'll be able to do that with hockey app? Because if we can slugify the branch name and allow people to install any version of the app they like from hockeyapp that would be incredible.

robbiemccorkell:
With the branch builds and automated testing we have now I don't think we really need this anymore. Going to close the PR for now.

PR #34, opened 2018-02-12
Android deploy

markholland:
We want to be able to deploy to testers/stakeholders as features are completed. We already have this in place for iOS, this PR is a base for doing the same for Android.

We've taken a similar approach to the Fastlane tools used for iOS by storing the Android keystore in a private repo. Currently the keystore being used is a self generated one. This will need to be swapped out or used as an `upload` keystore when we want to start getting builds into the Play Store.

Right now all we care about is getting builds to internal users. For this we have chosen to use HockeyApp which is free for upto 2 apps and allows a high number of testers.

This PR modifies our CI config to clone the private repo with our keystore and use it to build a signed `apk`. If the branch is master then we also post the signed `apk` to HockeyApp notifying all testers of the new build.

What has been left out for a future task is handling build & app versions. We need to decide how to ensure they stay the same across both Android & iOS and would lead to changes to the iOS pipeline.

### TODO
- [x] Clone private repo containing a keystore
- [x] Pass secrets as CircleCI env vars to gradle.properties for signed build
- [x] Post signed APK to hockey app
- [x] Commit message as release notes in HockeyApp

Before raising a pull request

* [ ] Documentation
* [ ] Unit tests

## Pre-merge check-list

* [ ] Tester approved


android/gradle.properties
```diff
@@ -10,7 +10,7 @@
 # Specifies the JVM arguments used for the daemon process.
 # The setting is particularly useful for tweaking memory settings.
 # Default value: -Xmx10248m -XX:MaxPermSize=256m
-# org.gradle.jvmargs=-Xmx2048m -XX:MaxPermSize=512m -XX:+HeapDumpOnOutOfMemoryError -Dfile.encoding=UTF-8
+org.gradle.jvmargs=-Xmx2048m -XX:MaxPermSize=512m -XX:+HeapDumpOnOutOfMemoryError -Dfile.encoding=UTF-8
```
markholland: This seemed to help with the bundleJS gradle task randomly failing quite often.

.circleci/config.yml
```diff
@@ -50,8 +50,25 @@ jobs:
           path: ~/project
       - attach_workspace:
           at: ~/project
+      - run: git clone git@github.com:redbadger/pride-android-secrets.git
+      - run: mkdir -p .gradle 
+      - run: echo -e "keystore=$ANDROID_KEYSTORE_PATH\nkeystore.alias=$ANDROID_KEYSTORE_ALIAS\nkeystore.password=$ANDROID_KEYSTORE_PASSWORD\nkeystore.keypass=$ANDROID_KEYSTORE_PASSWORD" >> .gradle/gradle.properties
       - run: bundle install
       - run: bundle exec fastlane build
```
frigus02: Does this do the Android build and signing?

markholland: Yes, we clone the keystore and then populate a properties file with the secrets and path of the keystore. Anyone can create their own properties file with their own keystore for local development. Then we use Fastlane(a Ruby gem) to handle the building of a signed binary.

robbiemccorkell:
So did it end up being easier to deploy to hockeyapp just in the circle config than through fastlane?

robbiemccorkell:
Also do you think we would do something else for a beta-android step, like go to the beta section of the Google play store?

robbiemccorkell:
And what do you want to do about the official keystore? Do we need someone (I think I can) to add the keystore to the shared place in Google play? And would you rather do that after this is merged?

markholland:
So this is just to make sure everyone is getting the master builds regardless of what device they have.

I do want to make all of this more CI agnostic so things will move to fastlane over time so that when we hand this project over it will be possible to continue to update and deploy the app with just a dev machine and not require a CI indefinitely.

Yes, so there's only an Alpha lane right now while there isn't much of interest to stakeholders. The plan is to have a Beta lane on both hockeyapp and TestFlight/Play Store and HockeyApp be more for badgers/volunteers and then the Stores for stakeholders.

We should get the official keystore uploaded asap but won't be used until the Beta lane is setup.

robbiemccorkell:
Got it ok. I might look at uploading the keystore later today.

That's a really interesting point about fastlane enabling builds off CI. Hadn't really thought about it that way before.

robbiemccorkell:
But yea happy to merge whenever you like 👍

PR #36, opened 2018-02-18
Pipeline improvements

markholland:
We want our own distribution of builds for internal QA.

Hockeyapp will provide all builds from all branches under the Alpha naming.

Builds can be distinguished by their notes in HockeyApp which contains the commit message.

Master builds will come under the Beta naming in HockeyApp and all testers will be notified. Master builds will continue to go to TestFlight and at some point Play Store Beta. That is where sign off and build promotion will happen.

App versioning is done with `npm version [major][.minor][.patch]`, this will update the `package.json` and be reflected in the native apps. A new commit and tag will also be created.

Distribution logic has been moved to the fastlane configs so as to be more CI agnostic.

Currently for Android both Alpha and Beta use the same signing config. This isn't possible on iOS so a third type known as `adhoc` is used for HockeyApp builds, this allows signing a binary for distribution outside of the App Store.

Beta builds match the production app in every way and the Beta builds published to TestFlight/Play Store Beta will be promoted to production for releases.

Alpha builds can be installed alongside Beta builds and can be distinguished by the `Alpha` suffix in the app name.

HockeyApp returns a URL to each build uploaded so potentially we could have a slack channel with build links for easy share with whoever is testing/reviewing.

After merging this we can add everyone to HockeyApp but there will be an extra step for iOS as every device has to be included in the provisioning profiles before builds can be installed, sigh.

- [x] Deploy iOS from master to HockeyApp 
- [x] Move HockeyApp logic to Fastlane
- [ ] Deploy Android from master to PlayStore Beta lane
- [x] Branch builds in HockeyApp installable alongside existing versions
- [x] Sync app version number across platforms

## Pre-merge check-list

* [ ] Tester approved


markholland:
The downside for HockeyApp on the iOS side is we have to manually add devices. I've added a lane for doing this but you'd still need the Apple credentials.

.flowconfig
```diff
@@ -19,6 +19,8 @@
 ; Ignore metro
 .*/node_modules/metro/.*
 
+.*/node_modules/config-chain/.*
```
markholland: Started getting flow errors without ignoring this

android/fastlane/README.md
```diff
@@ -0,0 +1,75 @@
+# fastlane documentation
```
markholland: Updating fastlane regenerates this

android/gradle.properties
```diff
@@ -10,7 +10,10 @@
 # Specifies the JVM arguments used for the daemon process.
 # The setting is particularly useful for tweaking memory settings.
 # Default value: -Xmx10248m -XX:MaxPermSize=256m
-org.gradle.jvmargs=-Xmx2048m -XX:MaxPermSize=512m -XX:+HeapDumpOnOutOfMemoryError -Dfile.encoding=UTF-8
+org.gradle.daemon=true
```
markholland: some recommended settings to improve build performance specifically with react native

package.json
```diff
@@ -10,7 +10,10 @@
     "test": "jest src",
     "precommit": "lint-staged",
     "lint": "eslint src",
-    "flow": "flow src"
+    "flow": "flow src",
+    "fix-xcode": "react-native-schemes-manager all",
```
markholland: This is for supporting adhoc builds in xcode with react native

markholland:
Some configs might not look very DRY but if alpha/beta starts to deviate more then it will be necessary.

ios/fastlane/Fastfile
```diff
@@ -13,34 +13,99 @@ platform :ios do
...
+      notes: notes,
+      status: options[:status],
+      notify: options[:notify],
+      commit_sha: options[:commit_sha]
+    )
+  end
+
+  desc "Add devices"
+  lane :devices do |options|
+    register_devices(
```
robbiemccorkell: Love that this is a lane too. At least it will be relatively easy to register new test devices.

robbiemccorkell:
Looks like the build status is getting confused above, not 100% sure why yet

markholland:
It’s because I renamed some workflow stages which are marked as expected by the repo. Will adjust prior to merging.

robbiemccorkell:
Ah ok. Might just have to force the PR through if needs be

markholland:
It's all good now, I didn't test the beta iOS lane as it triggers testflight so fingers crossed...

markholland:
Looks like beta iOS didn't work quite right so will have to look into that...

markholland:
Ok so I naively reused the adhoc build for Betas without thinking that the app name and logo would still be Alpha. Can add another config to pretty this up, build to TestFlight and ultimately production is still correct, but will give low priority.

PR #39, opened 2018-02-24
Add iOS device to adhoc profiles

markholland:
Provisioning profiles are per config so when adding a new device we need to regenerate both adhoc profiles as well.

## Pre-merge check-list

* [ ] Tester approved


markholland:
Although ready to merge it might make sense to wait for more people with iOS to be added then this merge will trigger a build they can all install

ios/fastlane/Fastfile
```diff
@@ -103,9 +103,22 @@ platform :ios do
 
   desc "Add devices"
   lane :devices do |options|
-    register_devices(
-      devices_file: "./devices.txt"
+    new_devices = get_unprovisioned_devices_from_hockey(app_bundle_id:'org.prideinlondon.festival', api_token:options[:api_token])
```
markholland: Need to dig in more with the fastlane plugins. This one is super useful!

markholland:
Not seeing anyone register their device so just going to merge this

PR #40, opened 2018-02-24
Add Redux

robbiemccorkell:
I was interested in pushing the project until we felt we needed redux to see how necessary it really was.

The more I thought about it, the more redux already feels useful:

- Syncing and fetching logic can be further abstracted away from pages.
- In memory data is easier to access from page to page due to the oddities of `react-navigation` api.
- With the use of actions and selectors, future work on sorting and filtering of events will be easy to abstract away
- There's a clear place to put logic for future pages and components.

## Pre-flight check-list

Before raising a pull request

* [x] Documentation
* [x] Unit tests

## Pre-merge check-list

* ~~Tester approved~~


.flowconfig
```diff
@@ -55,4 +55,4 @@ suppress_comment=\\(.\\|\n\\)*\\$FlowFixedInNextDeploy
 suppress_comment=\\(.\\|\n\\)*\\$FlowExpectedError
 
 [version]
-^0.63.0
+^0.65.0
```
robbiemccorkell: There is a more recent version of flow with far far better error descriptions, but currently it conflicts with some of the React Native flow types.

frigus02:
Do you think we should add an [ADR](https://github.com/redbadger/pride-london-app#architecture-decision-records) for this?

robbiemccorkell:
Good idea, I'd already forgotten about them. Done.

PR #42, opened 2018-02-25
Use Beta by Fabric for internal distirbution

markholland:
We want a simple way for people to test builds from feature branches.

We've already introduced the `alpha` flavor to allow installation alongside stable versions of the app. But when using HockeyApp as our dsitribution platform they only allow the latest build for iOS. This would cause issues if multiples features are being developed/tested concurrently.

Enter Beta by Fabric.

When using Fabric as an internal distribution platform it stores and makes available all builds for iOS. This means a tester would simply look through the list in Fabric on their phone and identify by the branch/commit-msg in the notes for each build.

Fabric was less favoured initially for two reasons:

- More difficult to setup and forces you to integrate their crash reporting system `Crashlytics`.
- Fabric was purchased from Twitter by Google and is clearly moving in a direction to be part of a larger stack to compete with Microsoft's App Center so its future was less certain than HockeyApp that has remained untouched by Microsoft since its purchase in 2014.

It only makes sense to migrate Android as well so that we are consistent in our distribution platform.
I'm in no rush to remove HockeyApp as we currently have a nice mechanism for getting peoples devices into our iOS Provisioning Profiles which would be used by Fabric as well.

## Pre-merge check-list

* [ ] Tester approved


.circleci/config.yml
```diff
@@ -116,6 +124,10 @@ jobs:
                 submit:true \
                 api_token:$HOCKEY_APP_TOKEN \
                 app_id:$HOCKEY_APP_IOS_ALPHA_APP_ID \
+                fabric_api_token:$FABRIC_API_KEY \
+                fabric_build_secret:$FABRIC_BUILD_SECRET \
+                fabric_groups:$FABRIC_GROUPS \
+                ci_build_num:$CIRCLE_BUILD_NUM \
```
markholland: Fabric ignores new builds with the same build number so for internal builds we'll just use the CI build number. On Android this is done through the gradle.properties file.

.circleci/config.yml
```diff
@@ -116,6 +124,10 @@ jobs:
                 submit:true \
                 api_token:$HOCKEY_APP_TOKEN \
                 app_id:$HOCKEY_APP_IOS_ALPHA_APP_ID \
+                fabric_api_token:$FABRIC_API_KEY \
+                fabric_build_secret:$FABRIC_BUILD_SECRET \
+                fabric_groups:$FABRIC_GROUPS \
```
markholland: Groups are a list of Fabric created groups of testers allowed to view the build in Fabric

android/fastlane/Fastfile
```diff
@@ -60,15 +76,34 @@ platform :android do
...
+      groups: options[:groups],
+      notifications: options[:notify]
+    )
+  end
+
+  def build_notes(commit_sha)
+    date = Time.now.strftime('%F')
+    time = Time.now.strftime('%T')
+    commit_msg = sh("git log --format=%B -n 1 #{commit_sha}").strip
+    notes = "#{git_branch}/#{commit_msg} on #{date} at #{time}"
```
markholland: Made some changes to build notes format, open to suggestions to help make it simple to identify a build for testing.

.circleci/config.yml
```diff
@@ -52,7 +52,8 @@ jobs:
           at: ~/project
       - run: git clone git@github.com:redbadger/pride-android-secrets.git
       - run: mkdir -p .gradle 
-      - run: echo -e "keystore=$ANDROID_KEYSTORE_PATH\nkeystore.alias=$ANDROID_KEYSTORE_ALIAS\nkeystore.password=$ANDROID_KEYSTORE_PASSWORD\nkeystore.keypass=$ANDROID_KEYSTORE_PASSWORD" >> .gradle/gradle.properties
+      - run: echo -e "keystore=$ANDROID_KEYSTORE_PATH\nkeystore.alias=$ANDROID_KEYSTORE_ALIAS\nkeystore.password=$ANDROID_KEYSTORE_PASSWORD\nkeystore.keypass=$ANDROID_KEYSTORE_PASSWORD\nversionCode=$CIRCLE_BUILD_NUM" >> .gradle/gradle.properties
+      - run: echo -e "apiSecret=$FABRIC_BUILD_SECRET\napiKey=$FABRIC_API_KEY" >> ./app/fabric.properties
```
markholland: Crashlytics needs the Fabric secrets

android/app/build.gradle
```diff
@@ -103,7 +103,10 @@ android {
         applicationId "com.pridelondonapp"
         minSdkVersion 16
         targetSdkVersion 22
-        versionName "1.0.3"
+        versionName "1.0.4"
+        Properties props = new Properties()
+        props.load(new FileInputStream(file("../.gradle/gradle.properties")))
+        versionCode props['versionCode'].toInteger()
```
markholland: This is the build number. Hasn't been needed for HockeyApp but Fabric only adds a new download if the build increments

ios/fastlane/Fastfile
```diff
@@ -62,13 +72,22 @@ platform :ios do
     if options[:submit]
       deploy_testflight 
     end
+    increment_build_number(build_number:options[:ci_build_num])
```
markholland: I don't see a reason to sync build numbers with those sent to TestFlight at this time.

package.json
```diff
@@ -14,7 +14,7 @@
     "flow": "flow src",
     "fix-xcode": "react-native-schemes-manager all",
     "postinstall": "yarn fix-xcode",
-    "postversion": "react-native-version"
+    "postversion": "react-native-version --reset-build"
```
markholland: Sets build number to one, not actually required with current setup but makes sense to do so.

markholland:
No rush on reviewing this, just wanted to get it working befor the week started

ios/PrideLondonApp.xcodeproj/project.pbxproj
```diff
@@ -1021,6 +1032,19 @@
...
+			buildActionMask = 2147483647;
+			files = (
+			);
+			inputPaths = (
+			);
+			outputPaths = (
+			);
+			runOnlyForDeploymentPostprocessing = 0;
+			shellPath = /bin/sh;
+			shellScript = "if [ \"${CONFIGURATION}\" != \"Debug\" ]; then\n    apiKey=$(sed -n '1p' < ./PrideLondonApp/fabric.properties)\n    apiSecret=$(sed -n '2p' < ./PrideLondonApp/fabric.properties)\n    /usr/libexec/PlistBuddy -c \"Set :Fabric:APIKey string $apiKey\" ./Info.plist\n    ./Fabric.framework/run $apiKey $apiSecret\nfi\n";
```
markholland: Here we build a plist with the fabric secrets

robbiemccorkell:
Very nice, looks like a good system. What do you think of Crashlytics? Is it something worth using or just an annoying dependancy?

markholland:
Crashlytics should be capable enough for our needs so probably makes sense to use it.

PR #43, opened 2018-02-28
Use Detox for greybox e2e testing

markholland:
Tried out Detox for on device e2e testing, only setup for iOS but Android is supported.

- https://github.com/wix/detox
- https://hackernoon.com/detox-gray-box-end-to-end-testing-framework-for-mobile-apps-196ccd9564ce

- [x] Make sure build is able to load from contentful i.e. write a test that interacts with an event.


e2e/init.js
```diff
@@ -0,0 +1,13 @@
+import detox from "detox";
+import { detox as config } from "../package.json";
+
+// Set the default test timeout of 120s
+jest.setTimeout(120000);
```
x4d3: out of curiosity, why 120 seconds ?

robbiemccorkell:
Detox looks like a pretty nice tool, hadn't seen it before. I like the fact it's all promise based. I hate it when e2e testing libraries try to be too clever.

e2e/init.js
markholland: Abitrary, e2e timeouts are usually quite high to allow for network latency.

markholland:
I'll clean this up today and get Android running as well as I think some basic e2e tests will be helpful.

e2e/init.js
```diff
@@ -0,0 +1,13 @@
+import detox from "detox";
+import { detox as config } from "../package.json";
```
charypar: What's the benefit of having detox config in `package.json`?

markholland: I could move it to its own config file, I'd put it in package.json because the jest config was already there.

markholland:
Considering master already has a regression, tapping on events does nothing, how do we feel about merging this with it only running on iOS for now? I think very basic manual testing can be quickly automated and kept up to date. The run time is currently similar to the iOS build so shouldn't cause any disruption.

robbiemccorkell:
Wow, that's a big regression I didn't realise. Yea sounds good, I think since we're much lighter on manual testing for this project, and automated golden path is even more valuable.

PR #47, opened 2018-03-06
[WIP] Event Details page

robbiemccorkell:
- [x] Fill in event details page
- [x] Add react-native-maps
- [x] Add fonts
- [ ] Fix all test data to add required fields

## Pre-flight check-list

Before raising a pull request

* [ ] Documentation
* [x] Unit tests

## Pre-merge check-list

* [ ] Tester approved


markholland:
Opps, misclick on closing.

Intrigued why adding fonts is such a native task? Especially as it creates these ugly conflicts.

robbiemccorkell:
Yea it's a bit odd @markholland. It seems that fonts are required to be added to the native layer on both android and ios. It's a much more complicated process for ios than android. It's all handled by `react-native link`. https://medium.com/react-native-training/react-native-custom-fonts-ccc9aacf9e5e

frigus02:
Do we really need to add the Roboto font to the Android app. I think it should already be available on all Android devices as it's the default system font. What am I missing?

robbiemccorkell:
@frigus02 You're not missing anything. The `react-native link` system adds our custom fonts to both platforms automatically. So we could go in and delete them again, but they would reappear each time someone runs `react-native link` again.

I could always add them to `.gitignore` I suppose

frigus02:
Okay. We should probably just leave it as it is then. Maybe a future version of React Native can deal with this automatically. Do they get included in the APK file during build?

robbiemccorkell:
I'm not sure. I would assume so, because for android it's just a folder of static files. I don't think it would know to exclude them.

src/components/Text.js
```diff
@@ -0,0 +1,101 @@
...
+
+type Props = {
+  children: Node,
+  type?: TextType,
+  markdown?: boolean,
+  style?: StyleObj
+};
+
+const Text = ({ children, type, markdown, style }: Props) =>
+  markdown ? (
```
frigus02: It looks like there is no functionality overlap between a simple text and a markdown view. The `type` prop is unused when `markdown` is `true`. Do you think it would make sense to split this into 2 separate components `<Text />` and `<MarkdownText />`?

src/constants/colors.js
```diff
@@ -7,3 +7,10 @@ export const textColor = "#272727";
 
 export const filterButtonColor = "#5C5C5C";
 export const filterButtontextColor = "#ADADAD";
+
+// EVENT DETAILS
+export const eventDetailsBgColor = "#fff";
+export const eventDetailsHeaderBgColor = "#F2F2F2";
+export const eventCategoryLabelBgColor = "#d8d8d8";
```
frigus02: Do we want to use uppercase characters for all colors?

src/constants/strings.js
```diff
@@ -0,0 +1,3 @@
+export default {
+  venuedetailsGenderNeutralToilets: "Gender neutral toilets"
```
frigus02: I think the "d" in "details" should be uppercase. 🔍

src/screens/EventDetailsScreen/Header.js
```diff
@@ -0,0 +1,38 @@
...
+type HeaderProps = {
+  onBackButtonPress: () => void
+};
+
+const Header = ({ onBackButtonPress }: HeaderProps) => (
+  <SafeAreaView style={styles.container}>
+    <StatusBar barStyle="default" animated />
+    <View style={styles.header}>
+      <TouchableOpacity onPress={onBackButtonPress}>
+        <Text>Back</Text>
```
frigus02: Should this text "Back" be extracted into `src/constants/text.js`?

src/screens/EventDetailsScreen/component.js
```diff
@@ -0,0 +1,225 @@
...
+  return (
+    <View style={styles.content}>
+      <Text type="h1">{event.fields.name[locale]}</Text>
+      <View style={styles.categoryLabelContainer}>
+        {event.fields.eventCategories[locale].map(categoryName => (
+          <CategoryLabel key={categoryName} categoryName={categoryName} />
+        ))}
+      </View>
+      <View style={styles.iconItemWrapper}>
+        <IconItem icon={<Text type="xSmall">icn</Text>} title={dateDisplay}>
```
frigus02: The usage of the `<IconItem>` is a bit confusing. I did not expect the `children` to be rendered below the `title`. I think one of these would be easier to understand:

```jsx
<IconItem icon={<Text type="xSmall">icn</Text>}>
  <Text>{dateDisplay}</Text>
  <Text type="small">{timeDisplay}</Text>
</IconItem>
```

```jsx
<IconItem
  icon={<Text type="xSmall">icn</Text>}
  title={dateDisplay}
  content={timeDisplay}
/>
```

src/screens/EventDetailsScreen/component.js
```diff
@@ -0,0 +1,225 @@
...
+  navigation: NavigationScreenProp<{ params: { eventId: String } }>,
+  event: Event
+};
+
+const renderEventOverview = event => {
+  const startTime = new Date(event.fields.startTime[locale]);
+  const endTime = new Date(event.fields.endTime[locale]);
+  const dateFormat = "DD MMMM YYYY";
+  const timeFormat = "HH:mm";
+  const dateDisplay = isSameDay(startTime, endTime)
```
frigus02: How does this work when the startTime and endTime are not on the same day? Example:

* Start: 2018-06-10 10:00
* End: 2018-06-12 18:00

In this case we would render: 2018-06-10 - 2018-06-12, 10:00 - 18:00.

This sounds like the event happens on three days (10, 11, 12) each from 10:00 to 18:00.

Is this how it is supposed to work?

src/selectors/events.js
```diff
@@ -3,3 +3,6 @@ const getEventsState = state => state.events;
 export const selectEvents = state => getEventsState(state).events;
 export const selectEventsLoading = state => getEventsState(state).loading;
 export const selectEventsRefreshing = state => getEventsState(state).refreshing;
+
+export const selectEventById = (state, id) =>
+  getEventsState(state).events.find(event => event.sys.id === id);
```
frigus02: We could use `selectEvents` here 🙂

src/constants/colors.js
robbiemccorkell: Yea that's cleaner, fixed.

src/constants/strings.js
robbiemccorkell: Good catch, fixed.

src/screens/EventDetailsScreen/Header.js
robbiemccorkell: Maybe, but I'm currently just using it as a placeholder for an Icon. The word `back` won't be there in the final design.

src/screens/EventDetailsScreen/component.js
robbiemccorkell: I think the second option is probably clearer. You're right I think I was trying to be too clever with the `children` node. I'll change that now...

robbiemccorkell: Done

src/screens/EventDetailsScreen/component.js
robbiemccorkell: I'm actually going to clarify this with UX because I think this current way doesn't make much sense.

I think you're assumption above is actually how it was intended to work, but I think there's many cases where it wouldn't make sense.

I think Steve has already changed this from the UX side so I'll clarify with him.

src/selectors/events.js
robbiemccorkell: Good point, fixed

src/components/Text.js
frigus02: > I felt this way might make it clearer that if you're changing the styles, you'll be changing the styles for both text and markdown.

That's a good argument. Agree.

My concern is mainly that the `type` prop is ignored when `markdown` is true. I think it's fine, though.

src/screens/EventDetailsScreen/Header.js
frigus02: Ohhh, that makes sense. We really don't need to extract placeholder text.

src/screens/EventDetailsScreen/component.js
frigus02: I wonder if `content` could even default to `<Text type="small">...</Text>`. 🤔

src/screens/EventDetailsScreen/component.js
```diff
@@ -0,0 +1,232 @@
...
+              text={text.eventDetailsBuyButton}
+              url={event.fields.ticketingUrl[locale]}
+            />
+          </View>
+        )}
+      </View>
+    </View>
+  );
+
+class EventDetailsScreen extends React.Component<Props> {
```
markholland: It's recommended to always use PureComponent for performance. Not sure when the best time to switch everyhting is though..

src/constants/strings.js
```diff
@@ -0,0 +1,3 @@
+export default {
```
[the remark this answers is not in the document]
robbiemccorkell: I'm using `text.js` for actual text displayed to the user, but this file contains a string that I am using in the code to find something from contentful, but it's not text that I would display to a user.

src/screens/EventDetailsScreen/component.js
robbiemccorkell: Fixed, there were only a couple of places it needed adding I think

PR #50, opened 2018-03-06
Fix local Android build

### LINKED ISSUE #44: Can't build locally for android simulator
Since this commit https://github.com/redbadger/pride-london-app/commit/0d61bf95bb69e2d79d5ad391d872197e2d600e8a `yarn run-android` stopped working. It throws the error `pride-london-app/android/.gradle/gradle.properties (No such file or directory)`.

I'm assuming the line `props.load(new FileInputStream(file("../.gradle/gradle.properties")))` is the culprit here. But I don't know enough about the android build system to know what it should be, or how to skip this step during local development. 

@markholland any ideas?

frigus02:
Fixes #44

* Add documentation about setting up the Android dev tools locally (Android Studio, Android SDK, Emulator).
* Fix `yarn run-android`. The main issue was the missing `--variant` parameter.
* Fix Crashlytics sometimes failing Android build.
* Add `dev` flavor for local development as default when running debug builds with `yarn run-android`.

## Pre-flight check-list

Before raising a pull request

* [ ] Documentation
* [ ] Unit tests

## Pre-merge check-list

* [ ] Tester approved


frigus02:
This works locally now. But something is wrong with the release build. I will continue with this tomorrow.

markholland:
@frigus02 https://github.com/react-navigation/react-navigation/issues/3097

frigus02:
Thanks @markholland for digging this up. I read through the issues. It seems to be an upstream thing (https://github.com/facebook/react-native/issues/16906), which was fixed here: https://github.com/facebook/react-native/commit/d16ff3bd8b92fa84a9007bf5ebedd8153e4c089d

I will check in the evening how we can use this. Unless I missed something this is not included in the latest React Native version.

markholland:
Is the fix for running Android locally dependent on upgrading gradle/android tools?

frigus02:
I don't think so. I ended up updating the build tools because I wanted to use Java 9. Unfortunetely it turned out Java 9 does not work at all. I might back the build tools update out.

Would be nice to have it, though. It has a few benefits:
- Faster builds
- Fonts as resources (this one could be useful)
- Java 8

frigus02:
It works now. I will check how this would work without the hack and without updating the build tools.

The new SDK would be so nice, though. We could build some really nice adaptive icon for the app 😃 
image

markholland:
The version number isn't being set anymore, this is important for distribution platforms to know whether it's new.

screen shot 2018-03-07 at 21 18 19

frigus02:
Ok. Seems like we need to specify the versionCode in the manifest file then.

frigus02:
Ok. Appearently the versionCode issue was introduced here: https://github.com/redbadger/pride-london-app/commit/18cea4f957c7533d4240e70875c04601c1616cde
```
versionCode props['version  Code'].toInteger()
                          ^^
```

We might as well include the fix in this PR.

android/fastlane/Fastfile
```diff
@@ -91,11 +91,12 @@ platform :android do
 
   desc "Upload to Beta by Fabric"
   lane :deploy_fabric do |options|
+    File.write("./notes.txt", options[:notes])
```
markholland: https://github.com/fastlane/fastlane/issues/6325#issuecomment-318026228

frigus02: Ahh good finding. This one was really annoying :+1:

markholland: Don’t know why it never happened before merging Crashlytics. Build is still flaky because it fetches android tools every time

frigus02: It's the Java dependencies, right? Downloading from `jcenter` seems flaky for some reason. We might be able to add these libraries to the Circle CI cache. This should make the build faster. And maybe less flaky.

frigus02:
@markholland I think this is good to go.

PR #52, opened 2018-03-09
CI caching android

markholland:
Caching Android and Ruby dependencies in CI

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -74,6 +74,8 @@ jobs:
     working_directory: ~/project/android
     docker:
       - image: circleci/android:api-27-node8-alpha
+    environment:
+      TERM: dumb
     steps:
       - checkout:
           path: ~/project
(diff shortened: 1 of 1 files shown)


markholland:
Pointing this at the fix android local branch as didn't want to experiment directly there but this should probably go with that to master.

markholland:
3m30s => 2m45s for the android-alpha job

.circleci/config.yml
```diff
@@ -45,16 +74,23 @@ jobs:
     working_directory: ~/project/android
     docker:
       - image: circleci/android:api-27-node8-alpha
+    environment:
+      TERM: dumb
```
frigus02: What does `TERM: dumb` do?

markholland: This CircleCI image doesn't set anything for TERM and gradle needs it to be set. I assume Fastlane sets it aswell but now that we're calling gradle directly earlier in the job this was needed. The value dumb is abitrary but recommended.

PR #56, opened 2018-03-10
Featured Events

StuartHarris:
https://trello.com/c/20qHCVLm/111-home-page-featured-events

image

- [x] Empty HomeScreen
- [x] Add "Featured Events" content model in Contentful, that points to between 4 and 6 events (for now)
- [x] UI from mocked data
- [x] Hook up to CMS data when available
- [x] Show all featured events when user taps on "View all"

* [x] Documentation
* [x] Unit tests
* [ ] Tester approved


robbiemccorkell:
This PR can now be integrated with the CMS properly once you pull from master. You can use the selector `selectFeaturedEvents` in `selectors/events.js`

robbiemccorkell:
Would be good to change the `intialRouteName` on our `TabNavigator` to the home page too in this PR.

src/screens/HomeScreen/component.js
```diff
@@ -0,0 +1,145 @@
...
+  onPress: (eventId: string) => void,
+  getAssetById: string => Asset
+};
+const Column = ({ column, events, onPress, getAssetById }: ColumnProps) => (
+  <View style={column === 0 ? styles.viewLeft : styles.viewRight}>
+    {events.map(event => (
+      <TouchableOpacity
+        key={event.sys.id}
+        style={styles.tileWrapper}
+        onPress={() => onPress(event.sys.id)}
```
frigus02: Sharing some things we learned from the event list:

In a scrolling view it is good to add `delayPressIn={50}` to the `TouchableOpacity`. Otherwise you will see the opacity effect, even when you are just scrolling. This is a bit distracting and does not really feel native. The delay helps.

I think we should add accessibility attributes to all custom buttons. Otherwise they are treated as text by screen readers. We need `accessibilityTraits` for iOS and `accessibilityComponentType` for Android. You can just copy the attributes from the `EventList` component:

https://github.com/redbadger/pride-london-app/blob/1e56cf838713589fcceb0abdf02ce85553f2e069/src/components/EventList.js#L33-L35

src/screens/HomeScreen/component.js
```diff
@@ -0,0 +1,128 @@
...
+      title: this.props.featuredEventsTitle
+    });
+  };
+
+  render() {
+    // Show only even number of events (2, 4 or 6).
+    // Never show more than 6 events.
+    const eventsCount = Math.min(
+      6,
+      this.props.featuredEvents.length - this.props.featuredEvents.length % 2
```
charypar: Might be more readable as `Math.floor(this.props.featuredEvents.length / 2)`

src/screens/HomeScreen/component.js
```diff
@@ -0,0 +1,128 @@
...
+            <Text>Header - TBD</Text>
+          </View>
+          <View style={styles.sectionTitle}>
+            <Text type="h2" style={{ color: titleTextColor }}>
+              {this.props.featuredEventsTitle}
+            </Text>
+            <TouchableOpacity onPress={this.eventList} testID="view-all">
+              <Text style={{ color: titleTextColor }}>View all</Text>
+            </TouchableOpacity>
+          </View>
```
charypar: I wonder whether this title should become it's own (sub)component, especially if we end up with more sections... Probably not a problem yet.

src/screens/HomeScreen/component.js
```diff
@@ -0,0 +1,128 @@
...
+                testID={`event-tile-${event.sys.id}`}
+              >
+                <EventTile
+                  name={event.fields.name[locale]}
+                  date={event.fields.startTime[locale]}
+                  imageUrl={`https:${
+                    this.props.getAssetById(
+                      event.fields.eventsListPicture[locale].sys.id
+                    ).fields.file[locale].url
+                  }`}
```
charypar: Maybe we should extract this into a helper function... `imageUrl(event.fields.eventsListPicture[locale])` or something

frigus02: Yeah, this is quite verbose currently and used in lots of components.

src/screens/HomeScreen/component.js
frigus02: Agree. I'm going to change this.

package.json
```diff
@@ -88,6 +88,9 @@
     "transformIgnorePatterns": [
       "node_modules/(?!(jest-)?react-native|react-navigation)"
     ],
+    "moduleNameMapper": {
+      "\\.(png|gif|jpeg|jpg)$": "RelativeImageStub"
+    },
```
frigus02: I ran this on Windows and it couldn't deal with imported images during unit tests. This seems to be the recommended fix.

frigus02:
I forgot accessibility attributes for the "View all" button. Will add them today. Then this is ready to merge.

PR #59, opened 2018-03-10
Event listing view all

trcywu:
## Pre-flight check-list

image

Before raising a pull request

* [ ] Documentation
* [ ] Unit tests

## Pre-merge check-list

* [ ] Tester approved


trcywu:
Hi @robbiemccorkell i'll be moving the formatting logic into eventCards but other than that can you sense check that this PR is going in the right direction? Thanks

src/components/EventList.js
```diff
@@ -23,16 +27,42 @@ const EventList = ({
...
+        timeFormat
+      )}`;
+      return (
+        <View style={styles.eventListItem}>
+          <TouchableOpacity
+            delayPressIn={50}
+            onPress={() => onPress(event.sys.id)}
+          >
+            <EventCard
+              date={dateDisplay}
```
charypar: Just a reminder of the conversation from yesterday: This prop should get replaced with two separate props: `startTime` and `endTime` which you can pass in the same way you do `name`. That change will allow you to move all of the date formatting logic into `EventCard` and make `renderItem` much simpler again.

The component got a whole lot more complicated in #62, so it will help a lot not to give it any more jobs.

PR #60, opened 2018-03-10
Handle all cms data offline

robbiemccorkell:
Save all cms data coming back from contentful and distribute to appropriate redux stores. Currently this will add ability to store:

- Other content models like featured events lists
- CMS assets like images

## Pre-flight check-list

Before raising a pull request

* ~~Documentation~~
* [x] Unit tests

## Pre-merge check-list

* ~~Tester approved~~


robbiemccorkell:
I'm almost done with this, but there's massive conflicts I need to sort out.

robbiemccorkell:
I think I'm done here.

The Events reducer now has an entries property which contains a mix of all content model types that we add (currently Event and FeaturedEvents). We also now have an assets property containing all assets.

src/integrations/storage.js
```diff
@@ -1,64 +1,76 @@
...
   syncToken: string
 };
 
-type SaveEvents = (
-  newEvents: Event[],
-  deletedEvents: Deletion[],
-  syncToken: string
-) => Promise<EventsData>;
-type LoadEvents = () => Promise<EventsData>;
+type CmsData = {
```
frigus02: Can you import this type from `cms.js`?

robbiemccorkell: Unfortunately not, due to a strange flow error. In this type I have to declare both asset arrays as `Object[]`. If I type them properly as `Asset[]` flow _freaks out_ and complains that `Asset` doesn't have the same fields as `Event`. I think it's something to do with the fact that both arrays are passed into the `updateCmsEntryList ` function and flow really doesn't like that. I've tried a bunch of different type signatures, and combinations of casting types to other things but without success.

So I ended up having to replicate this type, and `SavedData` in both `storage.js` and `cms.js`. The `cms.js` types are properly typed and exported to other modules, so at least that way the rest of the app gets proper typing for these arrays.

robbiemccorkell:
Yea absolutely. I can see the image React Native image components allow for indicators. Was hoping to ask design about it when we implement the full design on this page.

src/integrations/storage.js
frigus02: Ah; makes sense I guess...

Could you derive both `Event` and `Asset` from the same base type or interface, which has only the `sys.id` field? And then explicitly type `updateCmsEntryList` with the base type/interface? I could imagine, that flow will still have issues.

frigus02: Ah well. I guess it's fine.

robbiemccorkell: Yea I tried that too. I had a base `CmsEntry` type containing all sys metada with a generic type passed in representing the fields of this entry. Flow still complained that the types didn't have the same fields!

robbiemccorkell: This was one of the main things causing flow to drive me crazy and consider typescript. @paulroper had similar problems with flow on his last project.

frigus02: I just had a quick look. "flow *freaks* out" is definitely no overstatement 😄 
And you think TypeScript is able to deal with this sort of thing? Might be worth giving it a try.

robbiemccorkell: I can't say for sure that Typescript will not have the same problem. But we're using it on our current project and I've been pleasantly surprised. I've spent very little time debugging typescript errors on the project, and ages doing the same on this one.

PR #61, opened 2018-03-10
Filter Events List by Date and Time of Day

frigus02:
Add date and time filter to header.

Colors and spacing need to be polished, but we can do this in a separate PR.

iOS | Android
--- | ---
filter-ios | filter-android

## Pre-flight check-list

Before raising a pull request

* [x] Documentation
* [x] Unit tests

## Pre-merge check-list

* [ ] Tester approved


frigus02:
We might be able to use https://github.com/wix/react-native-calendars for the date picker view on iOS.

charypar:
While I'm still not convinced it's the right UX, this is some great work! Well done.

frigus02:
Yep, me neither. I wanted to know how we could get this working, though. I will focus on the filter screen and the actual filtering next.

frigus02:
I meant, I will focus on the date choosing screen (not filter screen) and the actual filtering. 😄

robbiemccorkell:
Awesome, nice one @frigus02. I like that you went with the action sheet as a compromise for iOS. At least it's the same number of taps for the user as Android that way.

src/components/DateRangePicker.js
```diff
@@ -0,0 +1,135 @@
...
+  markedDates[startDate] = {
+    ...template,
+    startingDay: true
+  };
+
+  // In between
+  let inBetweenDate = startDate;
+  do {
+    inBetweenDate = formatDate(addDays(inBetweenDate, 1), "YYYY-MM-DD");
+    markedDates[inBetweenDate] = { ...template };
```
charypar: If I'm reading it right, `markedDates` are immutable, so you could just reuse the `template` object here, without the spread. Not that it matters :)

src/components/DateRangePicker.js
```diff
@@ -0,0 +1,135 @@
...
+          ...dateRange,
+          endDate: day.dateString
+        })
+      );
+    }
+  };
+
+  render() {
+    const dateRange =
+      this.props.dateRange && typeof this.props.dateRange === "object"
```
charypar: Do we need to be this defensive? Could Flow help restricting the types of things we operate on?

frigus02:
The Code Climate issues are related to duplicated code in unit tests. I don't think this is really an issue here.

src/components/DateRangePicker.js
frigus02: We can absolutely reuse template here 🙂

src/components/DateRangePicker.js
frigus02: Hm, I couldn't get it working otherwise. You can have:
* `null`: No date selected, which means show all events.
* `string`: One date selected.
* a `DateRange` object: Two dates selected.

When I only check for `typeof this.props.dateRange === "object"` errors and says `dateRange` could be null.

frigus02:
I confirmed that the E2E tests work locally. This was just a CI issue.

PR #62, opened 2018-03-10
Group events by date

charypar:
Add grouping of events by date. I've implemented the data processing as a new selector.

I've also moved the Event type out into a separate file, which seems a lot cleaner, as we're working with it all over the place and pulling it in from the CMS integration module seems a bit wrong. 

I put that module into a `data` folder, which can be a sort of domain model section for functions that operate on the domain entities and the associated types.

## Screenshot

grouping

## Pre-flight check-list

Before raising a pull request

* [x] Documentation
* [x] Unit tests

## Pre-merge check-list

* [ ] Tester approved

### DIFF
--- a/src/components/EventList.js
+++ b/src/components/EventList.js
@@ -1,7 +1,7 @@
 // @flow
 import React from "react";
 import { StyleSheet, SectionList, TouchableOpacity, View } from "react-native";
-import { format, parse } from "date-fns";
+import { format } from "date-fns";
 
 import EventCard from "./EventCard";
 import Text from "./Text";
@@ -32,7 +32,7 @@ const EventList = ({
     keyExtractor={event => event.sys.id}
     sections={events.map(it => ({
       data: it,
-      title: format(parse(it[0].fields.startTime[locale]), "dddd D MMMM")
+      title: format(it[0].fields.startTime[locale], "dddd D MMMM")
     }))}
     renderSectionHeader={({ section }) => (
       <Text type="h2" style={styles.sectionHeader}>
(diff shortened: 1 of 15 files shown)


src/components/EventList.js
```diff
@@ -19,10 +27,20 @@ const EventList = ({
...
   onPress
 }: Props) => (
-  <FlatList
+  <SectionList
     contentContainerStyle={styles.container}
-    data={events}
     keyExtractor={event => event.sys.id}
+    sections={events.map(it => ({
+      data: it,
+      title: format(parse(it[0].fields.startTime[locale]), "dddd D MMMM")
```
frigus02: I don't think `parse` is necessay here. You can just pass a string to `format`.

src/selectors/events.js
```diff
@@ -1,5 +1,38 @@
...
+const locale = "en-GB";
+
+const groupByStartTime = (events: Event[]): EventDays => {
+  const sections = events
+    .sort((a: Event, b: Event) => {
+      const diff =
+        parseDate(a.fields.startTime[locale]) -
+        parseDate(b.fields.startTime[locale]);
+
+      return diff / Math.abs(diff);
```
frigus02: Why do we need to divide here?

charypar: The sorting callback expects `-1`, `0` or `1` back. This was a quick way to normalise if the difference is more than one day

src/components/EventList.js
charypar: Oh neat! I missed that. Convenient.

src/selectors/events.js
charypar: Ah, my bad, it doesn't care about the magnitude, probably thinking about a different language... Even better.

charypar:
@frigus02 Fixed the things you pointed out, also made some changes to the component to make it a bit more readable and added stronger types.

frigus02:
Looks good. Merge it. 😃 

(We should find a way to tell codeclimate that the unit tests are fine.)

PR #64, opened 2018-03-17
Display Price on Event List

trcywu:
## Pre-flight check-list

image

Pricing on price list will display 

1. Free
when isFree is true
2. From eventPriceLowest when there is a price range
3. eventPriceLowest when lowest === highest price

Before raising a pull request

* [ ] Documentation
* [ ] Unit tests

## Pre-merge check-list

* [ ] Tester approved

### DIFF
--- a/src/components/EventCard.js
+++ b/src/components/EventCard.js
@@ -16,7 +16,8 @@ type Props = {
   locationName: string,
   price: number,
   startTime: string,
-  endTime: string
+  endTime: string,
+  isFree: boolean
 };
 const removeTimezoneFromDateString = isoString => isoString.slice(0, -6);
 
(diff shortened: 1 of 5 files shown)


src/components/EventCard.js
```diff
@@ -14,18 +14,33 @@ import Text from "./Text";
...
+  eventPriceLow: number,
+  eventPriceHigh: number,
   startTime: string,
-  endTime: string
+  endTime: string,
+  isFree: boolean
 };
 const removeTimezoneFromDateString = isoString => isoString.slice(0, -6);
+const getEventPrice = (isFree, eventPriceLow, eventPriceHigh) => {
+  let displayPrice;
```
frigus02: You could get rid of the variable `displayPrice` and just return the strings from within the `if`.

PR #66, opened 2018-03-18
[Experiment] Hide filter header on scroll

frigus02:
## Pre-flight check-list

Before raising a pull request

* [ ] Documentation
* [ ] Unit tests

## Pre-merge check-list

* [ ] Tester approved

### DIFF
--- a/src/components/EventList.js
+++ b/src/components/EventList.js
@@ -1,9 +1,15 @@
 // @flow
 import React from "react";
-import { StyleSheet, SectionList, TouchableOpacity, View } from "react-native";
+import {
+  Animated,
+  StyleSheet,
+  SectionList,
+  TouchableOpacity,
+  View
+} from "react-native";
 import type { SectionBase } from "react-native/Libraries/Lists/SectionList";
+import type { Props as ScrollViewProps } from "react-native/Libraries/Components/ScrollView/ScrollView";
 import formatDate from "date-fns/format";
-
 import EventCard from "./EventCard";
 import Text from "./Text";
 import type { Event, EventDays } from "../data/event";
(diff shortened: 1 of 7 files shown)


frigus02:
I can't seem to get the refresh control working on iOS here. It is hidden behind the header and just won't move down. The suggestion I found on the internet is to use `contentInset` to move the whole list down and `contentOffset` to move the list items back up, but it keeps messing with the initial scroll of the list.

robbiemccorkell:
To be fair we might not actually have a refresh control in the final app, because it might suggest to users that they need to refresh a lot. Another question for UX/D. I imagine we might change the pull to refresh action, into just refreshing every time the app opens or something like that.

frigus02:
Ah, that's good to know. I think I will just park this here until we know from UX if and how scrolling should hide the header. I'm not even sure we want to have this at all, although I strongly vote for it. 🙂

robbiemccorkell:
If you come back to this any time soon, Jun mentioned that he would be interested in having the header shrink the other way on scroll, and show only the "I'm interested in" category filter banner while the user scrolls down.

frigus02:
Cool. I'll have a look next weekend.

JunBadger:
Having had a conversation with @cellyn-tan and @daralthus we think this feature would make a big difference to landscape orientation.

That said, the MVP for this will likely be a sticky header for portrait and non-stick for landscape. It also ties in to a bigger question of whether landscape is adding too much work for the user value. Should we be prioritising landscape against other stories? @andyc84

JunBadger:
Expected behaviour for Event Listing:
'Show me' and Top Date would always appear sticky
Only on scrolling up the secondary filters would appear

screen shot 2018-04-16 at 16 15 15

PR #68, opened 2018-03-21
Update content generation script

robbiemccorkell:
Our script needed updating to generate mock events since we now have far more fields than before.

This will allow us to much easier test filtering logic, and scrolling performance.

## Pre-flight check-list

Before raising a pull request

* [x] Documentation
* ~~Unit tests~~

## Pre-merge check-list

* ~~Tester approved~~


markholland:
Pretty miffed that e2e only got flaky after merging to master. I'll try to find time to address this. Also feel free to increase the function line limit in codeclimate, 25 is the default.

scripts/generate-content/contentful.js
```diff
@@ -0,0 +1,85 @@
...
+const MAX_EVENTS = 150;
+
+const generateEvents = (spaceId, accessToken, count) => {
+  const client = contentful.createClient({
+    accessToken
+  });
+
+  client
+    .getSpace(spaceId)
+    .then(async space => {
```
frigus02: Since we use `async`/`await` here already, I think could make the whole `generateEvents` function `async` and use a try-catch. Up to you, though 🙂

jonyardley:
Conflict 🍔

scripts/generate-content/contentful.js
robbiemccorkell: Good point, done

PR #69, opened 2018-03-24
Adding colours and styles to EvenList page

trcywu:
## Pre-flight check-list

image

Before raising a pull request

* [ ] Documentation
* [ ] Unit tests

## Pre-merge check-list

* [ ] Tester approved


src/components/Text.js
```diff
@@ -84,10 +73,16 @@ const textStyles = {
     fontSize: 14,
     lineHeight: 20
   },
-  xSmall: {
-    fontFamily: "Roboto",
+  price: {
```
robbiemccorkell: With the classes in this component I was kinda trying to follow the names of typographic styles laid out by the designers here: https://app.zeplin.io/project/5a916701f75c524181eebe68/screen/5ab1041c5727d6b45590a18d

It's changed since I last wrote these though. I feel like it might be neater to stick with generic classes like `h1`, `para`, `small` etc and then let components like `Price` that use this component just pick which typographic style it needs. If the styles need to be different from what's available you can pick the closest one and add extra styles through the normal style prop like `<Text style={{includeFontPadding: false}}>`.

It's not really a big deal, just felt conceptually neater to me

src/components/EventList.js
```diff
@@ -28,28 +30,28 @@ type ItemProps = { item: Event };
...
-        endTime={event.fields.endTime[locale]}
-        imageUrl={`https:${
-          getAssetById(event.fields.eventsListPicture[locale].sys.id).fields
-            .file[locale].url
-        }`}
-        isFree={event.fields.isFree[locale]}
-      />
-    </TouchableOpacity>
-  </View>
+  <TouchableOpacity
```
robbiemccorkell: I've started to notice that in most native apps the touchable elements darken when you touch them rather than lighten. I chose TouchableOpacity before I noticed this, but I wonder if we should use `TouchableHighlight` instead and try and find a nice value for it to darken the card when touched.

Doesn't have to change in this PR, just a thought.

robbiemccorkell:
I had a couple of comments that we can always follow up later. So pretty!

src/components/Text.js
frigus02: Yeah. We updated the styles here to match the ones in Zeplin again. Since the first designs are signed off now, I think hope also the text styles are mostly final.

src/components/EventList.js
frigus02: Yeah definitely. Also I would like to use [`TouchableNativeFeedback`](https://facebook.github.io/react-native/docs/touchablenativefeedback.html) on Android.

src/components/Text.js
frigus02: Also `price` and `tabBarItem` are now part of the Typography here: https://app.zeplin.io/project/5a916701f75c524181eebe68/screen/5ab1041c5727d6b45590a18d

trcywu:
@robbiemccorkell yes of course! sorry, wanted to get this PR in before lab tomorrow so that it didn't go stale

src/components/Text.js
robbiemccorkell: Oh I didn't see those. In that case ignore me then. If it matches the design guide I'd say that's fair.

robbiemccorkell:
No worries at all. There's nothing wrong in there. Just small suggestions.

PR #70, opened 2018-03-24
Add guide to reporting bugs to contributions.md

Midnite:
Could change that one link in case we'd rather track issues in Trello or somewhere else instead of github issues.


robbiemccorkell:
That `package-lock` file will have come from an `npm install`. If you delete it, and install dependencies with yarn in the future it won't appear.

Midnite:
@robbiemccorkell Oops, didn't notice that, deleted.

frigus02:
Nice work with the issue template. :+1:

The markdown headlines are a bit messed up. But instead of fixing this, maybe we want to extract this into a file `.github/ISSUE_TEMPLATE.md`. This way users will see it when they create a new issue.

Midnite:
@frigus02 good idea! will do it tomorrow

x4d3:
@frigus02 I've applied the requested change. (fyi @Midnite )

PR #74, opened 2018-03-25
Add instructions to run on a physical iPhone

x4d3:
Describe how to run the app on a physical iPhone


markholland:
This makes no sense for external contributors. Instructions should be for setting up their own free Apple dev account which is probably well documented in many places so a simple link would suffice.

PR #75, opened 2018-03-25
Initial event categories

RGBboy:
Add initial event category filtering functionality (no UI). This will be done by updating the the core state represented as a Set of categories selected. An empty set signifies no filter.

## Pre-flight check-list

Before raising a pull request

* [x] Documentation
* [x] Unit tests


RGBboy:
Had some flow issues so I had to amend. This is my first time using flow so if I have missed anything obvious, don't hesitate to call it out. In particular around optional named values for destructured arguments.

PR #77, opened 2018-03-25
Update Fastlane to latest version and set as minimum version

markholland:
Just keeping this updated as a lot of fixes come in each version.


jonyardley:
conflict 🍔

PR #79, opened 2018-03-25
Small improvements of date filter UX

charypar:
This improves the UX of the date picker filter dialog a little bit to be more intuitive. The main change is after selecting a date range, the next tap now starts a new date range. Also, tapping the grey backdrop will dismiss the dialog.

[add video]

The rest of this are little bugfixes.

## To do

- [x] make backdrop touchable to dismiss dialogs
- [x] more intuitive date range selection
- [x] fix visual bug caused by switching selection styles
- [ ] open date filter showing start of the date range
- [ ] when date filter is open with a single day selected, next tap should select a new single day (this will be tricky)

## Pre-flight check-list

Before raising a pull request

* [x] Unit tests

### DIFF
--- a/src/components/DateRangePicker.js
+++ b/src/components/DateRangePicker.js
@@ -27,8 +27,10 @@ const getSortedDateRange = (dates: DateRange) => {
 
 const getMarkedDate = (date: string) => ({
   [date]: {
-    selected: true,
-    selectedColor: eventListHeaderColor
+    startingDay: true,
+    endingDay: true,
+    color: eventListHeaderColor,
+    textColor: cardBgColor
   }
 });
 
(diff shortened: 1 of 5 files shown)


src/components/Dialog.js
```diff
@@ -73,6 +83,9 @@ Dialog.defaultProps = {
 };
 
 const styles = StyleSheet.create({
+  backdrop: {
+    ...StyleSheet.absoluteFillObject
```
frigus02: I think I made a mistake here using `absoluteFillObject`. I assume, that `flex: 1` would have the same effect. Would you mind testing this?

src/components/Dialog.js
```diff
@@ -39,30 +45,34 @@ const Dialog = ({
...
-            {headerLeft}
-          </View>
-          <Text type="h3" style={styles.headerTitle}>
-            {title}
-          </Text>
-          <View style={[styles.headerSide, styles.headerSideRight]}>
-            {headerRight}
+    <TouchableWithoutFeedback style={styles.backdrop} onPress={onCancel}>
+      <View style={styles.container}>
+        <TouchableWithoutFeedback>
```
frigus02: Why do we need 2 `TouchableWithoutFeedback` elements? Is this to only capture taps on the backdrop, but not on the dialog itself?

src/components/DateRangePicker.js
```diff
@@ -63,61 +60,58 @@ const getMarkedDateRange = (dateRange: DateRange) => {
...
-  onChange: DateOrDateRange => void,
-  dateRange?: ?DateOrDateRange
+  onChange: DateRange => void,
+  dateRange?: ?DateRange,
+  forceNewRange: boolean
 };
 
 class DateRangePicker extends React.PureComponent<Props> {
+  defaultProps: {
+    forceNewRange: false
```
frigus02: The prop is not optional. I don't think you need the default prop, do you?

PR #81, opened 2018-03-26
Standardize touchable components (e.g. buttons)

frigus02:
## Features

-   Enforce best practices for accessibility
    - Min touch area: 44x44
    - Accessibility attribute "button"
-   Use native feedback for Android
-   Offer all props of `TouchableWithoutFeedback` to make it easy to use in all places.

## Caveats

-   The native feedback for Android is hidden when views inside the button have a background color set.

### DIFF
--- a/src/components/Touchable.ios.js
+++ b/src/components/Touchable.ios.js
@@ -5,7 +5,7 @@ import type { TouchableProps } from "./TouchableTypes";
 import { TouchableDefaultProps } from "./TouchableTypes";
 
 const Touchable = ({ children, style, ...props }: TouchableProps) => (
-  <TouchableOpacity style={[styles.default, style]} {...props}>
+  <TouchableOpacity style={[styles.defaults, style]} {...props}>
     {children}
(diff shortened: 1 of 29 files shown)


src/components/Touchable.ios.js
```diff
@@ -0,0 +1,26 @@
...
+
+const Touchable = ({ children, style, ...props }: TouchableProps) => (
+  <TouchableOpacity style={[styles.default, style]} {...props}>
+    {children}
+  </TouchableOpacity>
+);
+
+Touchable.defaultProps = {
+  ...TouchableDefaultProps,
+  delayPressIn: 50
```
frigus02: Can someone test this on a physical iPhone and check if `delayPressIn` is necessary? How does scrolling in list views feel without this?

RGBboy: Had a look and this definitely makes a positive difference. With no delay, the press style is triggered almost every time when you try to start scrolling on a Touchable element.

PR #82, opened 2018-03-27
add instructions to get local device working for iOS

RGBboy:
Check out the readme! This needs to be verified with a new developer so ping me on slack to get the fastlane passphrase and test it out before we merge. I'm sure there are some amendments that will be needed.

- [ ] Verified with a new developer

### DIFF
--- a/README.md
+++ b/README.md
@@ -48,6 +48,26 @@ And install dependencies
 * Install Xcode from the App Store and accept the license agreement.
 * Run Xcode once so that it can install additional components it will need.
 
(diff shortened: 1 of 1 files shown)


declanslevin:
After running `fastlane match development --readonly`, cloning and decrypting the certificates repo, I was asked for the bundle identifier, which is `org.prideinlondon.festival`

jonyardley:
I can give this ago now. If I get the same message I'll add that and approve/mergé

jonyardley:
Needs a rebase @RGBboy

PR #85, opened 2018-03-28
104 & 127: Filter event tags

robbiemccorkell:
This PR covers both functionality and UI of this feature.

Functionality: https://trello.com/c/XPpWkEia/104-events-listing-event-tagging-filtering
UI: https://trello.com/c/I6flv9zS/127-ui-events-listing-event-tagging-filtering

Also contains some of the styling changes from https://github.com/redbadger/pride-london-app/issues/148 (but not all)

screen shot 2018-04-23 at 18 19 51
screen shot 2018-04-23 at 18 19 55

## Pre-flight check-list

Before raising a pull request

* ~~Documentation~~
* [x] Unit tests

## Pre-merge check-list

* [x] Tester approved

## Tester check-list

* [x] Tested on multiple devices / OS versions (Test on the physical device(s) you have access to, otherwise use BrowserStack):
  * [x] Galaxy S7 (Android 6.0)
  * [x] iPhone X (iOS 11.x)
  * [x] iPhone 8 (11.x)
  * [x] iPhone 7 (iOS 10.x)
  Optional:
    * [x] Google Pixel 2
    * [ ] Galaxy S8 Plus
    * [ ] Galaxy S7 Edge
    * [ ] Galaxy S6
    * [x] iPhone 7+
    * [x] iPhone 6

* [x] Text-to-Voice (Android: Voice Assistant / iOS: Speak Selection)
* [x] Large Text (=<200%)
* [x] Landscape Screen orientation

* [ ] Airplane mode
* [ ] 2G/3G Network Settings


src/components/EventCard.js
```diff
@@ -62,7 +62,9 @@ const EventCard = ({
             {name}
           </Text>
         </View>
-        <Text style={styles.eventLocation}>{locationName}</Text>
+        <Text type="small" style={styles.eventLocation}>
```
robbiemccorkell: This is unrelated to the story, but was some ad-hoc design feedback from Jun

src/components/ActionButton.js
```diff
@@ -0,0 +1,38 @@
...
+  onPress?: Function,
+  style?: StyleObj
+};
+
+const ActionButton = ({ label, onPress, style }: Props) => (
+  <Touchable
+    onPress={onPress}
+    style={style}
+    accessibilityTraits={["button"]}
+    accessibilityComponentType="button"
```
frigus02: Touchable defaults to button. So both accessibility attributes are not required.

src/components/ButtonPrimary.js
```diff
@@ -12,8 +17,13 @@ type Props = {
 };
 
 const Button = ({ children, disabled, onPress }: Props) => (
-  <Touchable style={styles.button} onPress={onPress} disabled={disabled}>
-    <Text type="h2" style={styles.text}>
+  <Touchable
+    style={[styles.button, disabled ? styles.disabledButton : {}]}
+    onPress={onPress}
+    disabled={disabled}
+    accessibilityTraits={disabled ? ["button", "disabled"] : ["button"]}
```
frigus02: It could be nice to move this logic into `Touchable.ios.js`. Adding the `disabled` could be generally useful.

src/components/CheckBox.js
```diff
@@ -5,6 +5,7 @@ import type { StyleObj } from "react-native/Libraries/StyleSheet/StyleSheetTypes
 import Text from "./Text";
 import Touchable from "./Touchable";
 
+import checkboxUrl from "./check-box.png";
```
frigus02: We have a new convention to put assets in `/assets/images`. Do you want to move these 2 images?

src/screens/FilterModal/component.js
```diff
@@ -0,0 +1,153 @@
...
+                        ? eventFilters[sectionName].size
+                        : null
+                    }
+                  />
+                  <ScreenSizeProvider>
+                    {size =>
+                      tags[sectionName].map(sectionValue => (
+                        <CheckBox
+                          key={sectionValue}
+                          onChange={() => {
```
charypar: Probably good idea to pull this out and reuse the function instance, mostly for clarity, but might also avoid some re-rendering

src/screens/FilterModal/Header.js
```diff
@@ -0,0 +1,57 @@
...
+        <Text
+          type="h2"
+          style={styles.titleText}
+          accessibilityTraits={["header"]}
+        >
+          {text.filterEvents}
+        </Text>
+      </View>
+      <ActionButton
+        key={showClear ? 1 : 0}
```
frigus02: What does the `key` do?

src/selectors/event.js
```diff
@@ -0,0 +1,9 @@
+// @flow
+import type { Event } from "../data/event";
+
+const locale = "en-GB";
```
frigus02: You can import this from `/src/data/locale.js`.

src/screens/FilterModal/component.js
```diff
@@ -0,0 +1,153 @@
...
+        forceInset={{ bottom: "always", top: "never" }}
+      >
+        <Header
+          onClearPress={this.clearTagFilters}
+          onCancelPress={() => navigation.goBack()}
+          showClear={numTagFiltersSelected > 0}
+        />
+        <ShadowedScrollView style={styles.flex} shadowOpacity={0.4}>
+          {Object.keys(tags).map(
+            sectionName =>
```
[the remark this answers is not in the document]
robbiemccorkell: The main reason for not using a SectionList was because Jun didn't want the sticky headers on this page. But I've only just noticed that `stickySectionHeadersEnabled` is a prop available on `SectionList`, so can use that.

src/screens/FilterModal/Header.js
robbiemccorkell: Strangely. I needed the key to make it re-render. But I've refactored the components since then so I'll double check it's needed.

src/selectors/event.js
robbiemccorkell: Shows how old this PR is...

src/components/ContentPadding.js
```diff
@@ -44,4 +46,5 @@ ContentPadding.defaultProps = {
   style: {}
 };
 
+export const getContentPadding = getPaddingStyle;
```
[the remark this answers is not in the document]
robbiemccorkell: Possibly, just got to find a decent place for it

src/components/ContentPadding.js
```diff
@@ -23,7 +25,7 @@ const defaultPadding = {
   large: { horizontal: 0, vertical: 0 }
 };
 
-const getPaddingStyle = (padding, size: ScreenSize) => {
+const getPaddingStyle = (size: ScreenSize, padding?: PaddingOption) => {
```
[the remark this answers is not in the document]
robbiemccorkell: Which part do you mean? The function name?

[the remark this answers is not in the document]
charypar: Yep

src/screens/FilterModal/component.js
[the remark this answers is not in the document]
charypar: Oh, interesting, I thought those would be a nice UX bonus :) oh well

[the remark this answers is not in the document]
charypar: ...maybe the visual design should then be closer to this kind of thing: 

. 

(I think the only real difference is the alignment of text in the gap)

src/screens/FilterModal/Header.js
robbiemccorkell: Yea I've checked this again and strangely the component does not re-render when the style prop changes. Changing the key forces the renderer to take notice of the change. I'll add a comment in the code to make it clearer.

src/components/ContentPadding.js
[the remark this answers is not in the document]
robbiemccorkell: I've been thinking about this more, and I'm not sure it's such a bad thing. The logic really belongs to the component, it's just in this instance we need the logic and can't use the component directly.

The alternative is that it moves into some kind of utils/lib file. I think this way it's at least close to the component that uses it most of the time.

src/screens/FilterModal/component.js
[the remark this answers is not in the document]
robbiemccorkell: Ok, I've broken up this into a couple of smaller components now, `FilterSectionList`, and `FilterList`. Looks cleaner and made the tests nicer too.

I looked into turning this into a `SectionList`, but I think it's going to turn quite messy and complicated putting it inside `ShadowedScrollView` as we'll have nested scroll views. I think it would require quite a big refactor of `ShadowedScrollView` to allow us to use its animated shadows around other elements. Luckily we don't really need the performance benefit here of the `SectionList` because it's such a short list. So I think it might be best to leave it as-is for now until the UX changes again.

robbiemccorkell:
@charypar @frigus02 I've worked on the comments above. Done almost all of the changes. As explained above, I've refactored the main Filter Modal component quite a bit, but haven't turned it into a SectionList

robbiemccorkell:
Also as a bonus, I've ended up strongly typing all the string set filters (the flow gods were kind to me this time), and bumped up the test coverage a little to pass codeclimate.

src/components/Touchable.ios.js
```diff
@@ -5,7 +5,11 @@ import type { TouchableProps } from "./TouchableTypes";
 import { TouchableDefaultProps } from "./TouchableTypes";
 
 const Touchable = ({ children, style, ...props }: TouchableProps) => (
-  <TouchableOpacity style={[styles.defaults, style]} {...props}>
+  <TouchableOpacity
+    style={[styles.defaults, style]}
+    {...props}
+    accessibilityTraits={props.disabled ? ["button", "disabled"] : ["button"]}
```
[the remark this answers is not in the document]
robbiemccorkell: Oh good point I'll do that now

robbiemccorkell:
Updated the checbox assets to include the proper sizes. They also have transparent backgrounds so android touch feedback looks better

src/data/event-filters.js
```diff
@@ -1,19 +1,46 @@
...
+  | Set<VenueDetail>
+  | Set<AccessibilityOption>
+  | Set<Time>;
+
+export type StringFilterSet =
+  | Set<EventCategoryName>
+  | Set<Audience>
+  | Set<VenueDetail>
+  | Set<AccessibilityOption>;
+
```
charypar: Nice! 👍

JunBadger:
@robbiemccorkell  - can we check the 'Select dates' / 'Add filters' is centred from the centre of the screen to the edge of the screen?
Currently looks like they're centred to the edge of the grid.

robbiemccorkell:
@JunBadger You're absolutely right, it's centred from the edge of the grid. I read that wrong from the image. Will change when I get back to it this evening.

JunBadger:
Thanks, gave it a check and looks good! @robbiemccorkell

PR #101, opened 2018-04-06
124 Filter events by categories

daralthus:
## Summary

Adds the category selection screen, where users can select which categories they want the events filtered by. 

[Trello card](https://trello.com/c/Qaxek2qR/124-ui-events-listing-filter-events-list-by-categories)

## Long description

- As user makes category selection, the CTA will update with the number of events ( to read: "Show X events")
- “Clear All” CTA appears when user makes one or more selection in the category list
- Clicking “Clear all” strips all selections from the list
- Clicking “Cancel” reverts back to the previous selection state (if user had previously made a category selection) and returns the user to the events listing
- If a user selects multiple categories (e.g. nightlife & music), the count in CTA will calculate towards events tagged with 'music' AND/ OR 'nightlife', i.e.they will not be mutually exclusive
- Only in the scenario where that there are no more events within a category, e.g. it is the last week of the festival and there are no more ‘health’ events then the health category should be removed from the filter list
- If there are no more events in a category because a user has already selected OTHER filter options (e.g. accessibility requirements) then the CTA should be disabled
- As a user selects a category it is displayed at the top of the page (in the “show me” box) in the order that it has been selected.

## Ticket / Screens

https://app.zeplin.io/project/5a916701f75c524181eebe68/screen/5ab0408d372ffa6b20773e0e
https://trello.com/c/Qaxek2qR/124-ui-events-listing-filter-events-list-by-categories

# Design changes
- `Show me` is hidden if font scale is above 1.5
- buttons and pills have been scaled to 44 instead of 40

## Pre-flight check-list

Before raising a pull request

* [x] Works on iOS
* [x] Works on Android
* [x] Documentation
* [x] Unit tests

## Pre-merge check-list

* [ ] Tester approved


.flowconfig
```diff
@@ -2,6 +2,9 @@
 ; We fork some components by platform
 .*/node_modules/.*/*[.]android.js
 
+# Ignore broken react-native-linear-gradient prop annotations
+.*/node_modules/react-native-linear-gradient/*
```
daralthus: Their prop type declarations seem broken with react 16 so had to opt out.

android/app/build.gradle
```diff
@@ -223,6 +223,7 @@ repositories {
 
 dependencies {
     compile project(':react-native-splash-screen')
+    compile project(':react-native-linear-gradient')
```
daralthus: Drop shadows are only supported with `elevation` on android and you can't set the direction of the light so gradients it is.

src/components/CategoriesPills.js
```diff
@@ -0,0 +1,144 @@
...
+import eventCategories from "../constants/event-categories";
+import {
+  blackColor,
+  whiteColor,
+  coralColor,
+  darkBlueGreyTwoColor,
+  eucalyptusGreenColor
+} from "../constants/colors";
+
+const locale = "en-GB";
```
daralthus: Localization has been considered in the components, but needs further work to fully integrate.

src/components/CategoriesPills.js
```diff
@@ -0,0 +1,144 @@
...
+import Text from "./Text";
+import text from "../constants/text";
+import eventCategories from "../constants/event-categories";
+import {
+  blackColor,
+  whiteColor,
+  coralColor,
+  darkBlueGreyTwoColor,
+  eucalyptusGreenColor
+} from "../constants/colors";
```
daralthus: Colors from the Zeplin style guide ftw! ❤️  💙 💚

src/screens/CategoriesFilterScreen/ListItem.js
```diff
@@ -0,0 +1,140 @@
...
+
+  constructor(props: ListItemProps) {
+    super(props);
+
+    // No initial animation
+    Animated.timing(this.decorationWidth, {
+      toValue: Number(props.selected),
+      duration: 0,
+      useNativeDriver: true
+    }).start();
```
daralthus: Showing the screen with selected categories won't animate the backgrounds.

src/components/CategoriesPills.js
joaquimadraz: I guess @jonyardley did that already https://github.com/redbadger/pride-london-app/pull/76

src/components/ShadowedScrollView.js
```diff
@@ -0,0 +1,149 @@
...
+    this.topShadowOpacity = new Animated.Value(0);
+    this.bottomShadowOpacity = new Animated.Value(1);
+  }
+
+  handleScroll = (event: { nativeEvent: { contentOffset: { y: number } } }) => {
+    const maxScrollOffset = this.contentViewHeight - this.scrollViewHeight - 10;
+    const currentScrollOffset = event.nativeEvent.contentOffset.y;
+
+    if (this.isTopShadowPresent && currentScrollOffset <= 0) {
+      this.isTopShadowPresent = false;
```
joaquimadraz: This flag is needed to avoid triggering the `fadeTopShadow` multiple times as it slows the animation.
Also it's not on the component state to avoid triggering the component render.

src/components/ShadowedScrollView.js
```diff
@@ -0,0 +1,149 @@
...
+      duration: shadowFadeDuration,
+      useNativeDriver: true
+    }).start();
+
+  componentWillMount() {
+    this.topShadowOpacity = new Animated.Value(0);
+    this.bottomShadowOpacity = new Animated.Value(1);
+  }
+
+  handleScroll = (event: { nativeEvent: { contentOffset: { y: number } } }) => {
```
joaquimadraz: Depending on the scroll offset, show or not the shadows indicating that there's more content 👍

src/constants/event-categories.js
```diff
@@ -1,41 +1,58 @@
...
+      color: colors.warmPinkColor
+    },
+    "Social and Networking": {
+      label: "Social and Networking",
+      color: colors.turquoiseBlueColor
+    },
+    Nightlife: {
+      label: "Nightlife",
+      color: colors.yellowColor,
+      contrast: true
```
joaquimadraz: This flag allows us to controller the colors for accessibility.
For example, the `Nightlife` category:

screen shot 2018-04-06 at 17 23 49

src/screens/EventsScreen/index.js
```diff
@@ -18,19 +18,14 @@ type OwnProps = {
...
 };
 
-type Props = {
-  events: EventDays,
-  loading: boolean,
-  refreshing: boolean,
-  updateEvents: () => Promise<void>,
-  getAssetUrl: LocalizedFieldRef => string
-} & OwnProps;
+type Props = ComponentProps & OwnProps;
```
joaquimadraz: These props where being duplicated from the component props. @daralthus came with this to reuse them.

android/app/build.gradle
frigus02: This should be `implementation` instead of `compile`. We can also fix it for the splash screen dependency above.

Android used to use `compile` but switched to `implementation` with one of the last build tools updates. React Native is not up-to-date yet and generates the old style.

src/components/CategoriesPills.js
```diff
@@ -0,0 +1,144 @@
...
+      </View>
+    );
+  }
+}
+
+const styles = StyleSheet.create({
+  selectedCategoriesPills: {
+    backgroundColor: darkBlueGreyTwoColor,
+    flex: 1,
+    height: 40,
```
frigus02: This should be `minHeight` instead. We need to allow the user to scale the font size up to 200%. If doing so, the pills need to get higher, otherwise the text is cropped.

For the shadow we need to find a way to use 100% of the height then, instead of fixed 40.

src/components/FilterHeader.js
```diff
@@ -27,6 +25,10 @@ type State = {
 };
 
 class FilterHeader extends React.PureComponent<Props, State> {
+  static defaultProps = {
+    onFilterCategoriesPress: () => {}
```
frigus02: Do we really need a default prop for this? In the `Props` it looks like it is required.

src/components/FilterHeaderCategories.js
```diff
@@ -0,0 +1,112 @@
...
+type CategoriesFilterButtonProps = {
+  style?: StyleObj,
+  onPress: Function
+};
+
+const CategoriesFilterButton = ({
+  style,
+  onPress
+}: CategoriesFilterButtonProps) => (
+  <TouchableHighlight
```
frigus02: Can you use our `Touchable` component here? It uses the native highlighting on Android, has the button accessibility trait by default and enforces a minimum size for the touch area.

src/components/FilterHeaderCategories.js
```diff
@@ -0,0 +1,112 @@
...
+    color: filterShowMeTextColor,
+    paddingTop: 5,
+    marginRight: 8
+  },
+  categoryPillsContainer: {
+    flexDirection: "row"
+  },
+  categoriesFilterButton: {
+    width: 38,
+    height: 40,
```
frigus02: Because we can't use a fixed height for the `interestButton`, this might need a height of 100%? Also it should have a minimal width of 44, which is the minimal size for touch areas for accessibility reasons.

I think I saw something about making only the touch area larger, but not the view itself. This might help us to still use the designed sizes.

src/components/FilterHeaderCategories.js
```diff
@@ -0,0 +1,112 @@
...
+    justifyContent: "center",
+    alignItems: "center",
+    borderTopRightRadius: 4,
+    borderBottomRightRadius: 4
+  },
+  categoryPills: {
+    borderTopRightRadius: 0,
+    borderBottomRightRadius: 0
+  },
+  categoriesFilterOverButton: {},
```
frigus02: Looks like this empty style can be removed 🙂

src/components/FilterHeaderCategories.js
```diff
@@ -0,0 +1,112 @@
...
+  interestButton: {
+    width: 145,
+    height: 40,
```
frigus02: We can't use fixed width and height here, because we need to allow users to scale the font. I think `minHeight` should work fine. And maybe a `width` is not needed at all?

src/components/ShadowedScrollView.js
```diff
@@ -0,0 +1,149 @@
...
+      opacity: this.bottomShadowOpacity.interpolate({
+        inputRange: [0, 1],
+        outputRange: [0, 1]
+      })
+    };
+
+    return (
+      <View style={style}>
+        <Animated.View style={[styles.topShadow, topShadowOpacityStyle]}>
+          <LinearGradient
```
frigus02: I wonder if you could make the `LinearGradient` itself animated and get rid of the wrapper view by using `Animated.createAnimatedComponent(LinearGradient)`. 🤔

src/screens/CategoriesFilterScreen/Header.js
```diff
@@ -0,0 +1,79 @@
...
+  selectedCategories: Set<string>
+};
+
+type ActionButtonProps = {
+  label: string,
+  onPress?: Function
+};
+
+const ActionButton = ({ label, onPress }: ActionButtonProps) => (
+  <TouchableOpacity style={styles.actionButton} onPress={onPress}>
```
frigus02: Can you use our `Touchable` component here to get the native styling on Android and the accessibility defaults?

src/screens/CategoriesFilterScreen/Header.js
```diff
@@ -0,0 +1,79 @@
...
+
+const styles = StyleSheet.create({
+  actionButtons: {
+    width: "100%",
+    flexDirection: "row",
+    justifyContent: "space-between",
+    marginTop: 10
+  },
+  actionButton: {
+    width: 60
```
frigus02: I think this should be `minWidth` to allow for font scaling.

src/screens/CategoriesFilterScreen/Header.js
```diff
@@ -0,0 +1,79 @@
...
+  },
+  actionButtonText: {
+    color: whiteColor
+  },
+  categoriesPills: {
+    width: "100%",
+    height: 40,
```
frigus02: Should be `minHeight` for font scaling 🙂

src/screens/CategoriesFilterScreen/ListItem.js
```diff
@@ -0,0 +1,140 @@
...
+
+type ListItemProps = {
+  category: EventCategory,
+  onPress: Function,
+  selected: boolean
+};
+
+class ListItem extends Component<ListItemProps, { textWidth: number }> {
+  static defaultProps = {
+    onPress: () => {}
```
frigus02: Do we need this default prop?

src/screens/CategoriesFilterScreen/ListItem.js
```diff
@@ -0,0 +1,140 @@
...
+  };
+
+  constructor(props: ListItemProps) {
+    super(props);
+
+    // No initial animation
```
frigus02: This comment confuses me slightly. No animation ... lets start an animation 😄 
It's because of the `duration` set to 0, right?

src/screens/CategoriesFilterScreen/ListItem.js
```diff
@@ -0,0 +1,140 @@
...
+  handleOnLayout = (event: { nativeEvent: { layout: { width: number } } }) =>
+    this.setState({ textWidth: event.nativeEvent.layout.width });
+
+  render() {
+    const { category, onPress, selected } = this.props;
+    const { textWidth } = this.state;
+    const textColor = selected && category.contrast ? blackColor : whiteColor;
+
+    return (
+      <TouchableOpacity
```
frigus02: Can we use our `Touchable` component here for accessibility defaults and native styling on Android?

src/screens/CategoriesFilterScreen/ListItem.js
```diff
@@ -0,0 +1,140 @@
...
+    height: 48,
+    width: Dimensions.get("window").width
+  },
+  check: {
+    position: "absolute",
+    marginTop: 16,
+    marginLeft: 6
+  },
+  itemText: {
+    height: 48,
```
frigus02: `minHeight` because of font scaling.

src/screens/CategoriesFilterScreen/component.js
```diff
@@ -0,0 +1,123 @@
...
+import locale from "../../data/locale";
+
+type Props = {
+  navigation: NavigationScreenProp<NavigationState>,
+  events: Event[],
+  stagedCategories: Set<string>,
+  toggleCategoryFilter: Function,
+  onApplyFilters: Function,
+  onClearAll: Function,
+  onClose: Function
```
frigus02: If you know the signature of the function, you can tell flow: `onClose: () => {}`

src/screens/CategoriesFilterScreen/component.js
```diff
@@ -0,0 +1,123 @@
...
+  static navigationOptions = {
+    header: null,
+    tabBarVisible: false
+  };
+
+  static defaultProps = {
+    toggleCategoryFilter: () => {},
+    onApplyFilters: () => {},
+    onClearAll: () => {},
+    onClose: () => {}
```
frigus02: Do we need these default props?

src/screens/CategoriesFilterScreen/component.js
```diff
@@ -0,0 +1,123 @@
...
+        <View style={styles.list}>
+          <List
+            locale={locale}
+            stagedCategories={stagedCategories}
+            onPress={this.handleFilterChange}
+          />
+        </View>
+        <ContentPadding>
+          <View>
+            <TouchableOpacity
```
frigus02: Use our `Touchable` component for native Android styling and accessibility defaults. 🙂

src/components/ShadowedScrollView.js
daralthus: It was using that originally, but I remember @joaquimadraz had some issues with it on iOS. I will try to replicate it.

src/components/FilterHeaderCategories.js
daralthus: According to https://facebook.github.io/react-native/docs/view.html#hitslop the touch area would not extend outside of the parent, so we need to bump the design to 44.

src/components/ShadowedScrollView.js
frigus02: Well, that's fine for me. You can leave it like this. Up to you 🙂

src/components/FilterHeaderCategories.js
frigus02: Ah, so there's a catch. So if UX wants this to be of height 40, we can still use our `Touchable` component and override the `minHeight` for it.

src/screens/CategoriesFilterScreen/ListItem.js
daralthus: Right, but it was trying to point out the reason why it's set to 0. Long story short, the `selected` state is coming through props so it would animate all selected categories when you reopen the panel. There's no need for that hence the code & comment.

src/components/FilterHeaderCategories.js
daralthus: Just updated the design to 44 and had a review with ux.

frigus02: Awesome :+1:

daralthus:
## Quick update

Android is getting there
image
image
image


but there are still some issues on iOS with the category pills component:

image
Will probably need to rely on something like this
https://github.com/CMP-Studio/TheWarholOutLoud/blob/8ae8ce688ad827118f3c1ab444e3b5763cafd1fa/ios/CMSAccessibilityManager.swift and this https://facebook.github.io/react-native/docs/pixelratio.html#getfontscale to make this kind of work less painful. What do you think @frigus02 ?

frigus02:
Hm, so what exactly is the issue on iOS? I hoped that using `minHeight` instead of `height` just solves everything. Maybe that was naive 😄 

getFontScale only works on Android, right?

daralthus:
minHeight ends up like this on iOS
image

`getFontScale` is android only, but I think we can access the font scale with a native module on iOS

daralthus:
Ok turns out getFontScale works quite well on iOS too. image

src/components/CategoriesPills.js
```diff
@@ -0,0 +1,144 @@
...
+    );
+  }
+}
+
+const styles = StyleSheet.create({
+  selectedCategoriesPills: {
+    backgroundColor: darkBlueGreyTwoColor,
+    flex: 1,
+    alignItems: "center",
+    height: 44 * PixelRatio.getFontScale(),
```
frigus02: That's actually a nice solution. Looks very clean.

I wanted to be picky and say: "How does this work when you change the font size while the app is open?".

But then I tried it. Looks like React Native just restarts the whole app in this case, even resetting the navigation. Interesting. But makes our live easier here.

src/components/Touchable.ios.js
```diff
@@ -5,7 +5,12 @@ import type { TouchableProps } from "./TouchableTypes";
 import { TouchableDefaultProps } from "./TouchableTypes";
 
 const Touchable = ({ children, style, ...props }: TouchableProps) => (
-  <TouchableOpacity style={[styles.defaults, style]} {...props}>
+  <TouchableOpacity
+    accessibilityTraits={["button"]}
+    accessibilityComponentType="button"
```
frigus02: Both `accessibilityTraits` and `accessibilityComponentType` are part of the default props (see the `TouchableDefaultProps` import). So they should not be needed here.

I tested this in the accessibility inspector and it works fine. Did you have any issues without these explicit props?

src/components/Touchable.android.js
```diff
@@ -4,12 +4,23 @@ import { StyleSheet, TouchableNativeFeedback, View } from "react-native";
...
+const Touchable = ({
+  children,
+  style,
+  onPress,
+  disabled,
+  ...props
+}: TouchableProps) => (
   <TouchableNativeFeedback
+    accessibilityTraits={["button"]}
+    accessibilityComponentType="button"
```
frigus02: Both `accessibilityTraits` and `accessibilityComponentType` are part of the default props (see the `TouchableDefaultProps` import). So they should not be needed here.

In fact, they cannot be here. This makes it impossible to override them. Some touchable components might be checkboxes, which have different traits.

src/components/Touchable.android.js
```diff
@@ -4,12 +4,23 @@ import { StyleSheet, TouchableNativeFeedback, View } from "react-native";
...
     background={TouchableNativeFeedback.SelectableBackground()}
-    {...props}
+    onPress={onPress}
+    disabled={disabled}
   >
-    <View style={[styles.defaults, style]}>{children}</View>
+    <View style={[styles.defaults, style]} {...props}>
```
frigus02: The props currently contain all properties the `TouchableNativeFeedback` can handle. It's a bit weird to apply them to the view instead.

If there is a need to put extra props on the nested view, we might need to add a special prop for this. Or rethink the automatically added nested view here. I put it there because there are restrictions for the children of the `TouchableNativeFeedback`:
> At the moment it only supports having a single View instance as a child node, as it's implemented by replacing that View with another instance of RCTView node with some additional properties set.

In any case, I think these props should be applied to the `TouchableNativeFeedback`.

src/components/CategoriesPills.js
```diff
@@ -0,0 +1,144 @@
...
+          <Text type="h3" style={styles.zeroSelected}>
+            {text.zeroSelected}
+          </Text>
+        ) : (
+          <View style={styles.scrollView}>
+            <LinearGradient
+              style={[styles.scrollShadow, styles.scrollShadowLeft]}
+              end={{ x: 0, y: 0.5 }}
+              start={{ x: 1.0, y: 0.5 }}
+              colors={["rgba(0, 0, 0, 0)", "rgba(0, 0, 0, 0.3)"]}
```
RGBboy: If we start doing more gradient overlays I suggest moving these to a common place.

src/components/CategoriesPills.test.js
```diff
@@ -0,0 +1,58 @@
...
+    const pillCustomStyles = output
+      .find("CategoryPill")
+      .dive()
+      .prop("style")[1];
+
+    expect(pillCustomStyles.backgroundColor).toEqual(expectedColor);
+  });
+
+  describe("#scrollToLastPill", () => {
+    it("scrolls to the end, if there is a scroll view", () => {
```
RGBboy: Not that this is a big deal, but it feels strange if you are scrolled to the start and remove the first category especially if a number of categories are selected. Would be good to get design/UX input here.

src/data/event.js
```diff
@@ -77,3 +77,11 @@ export type Asset = {
 };
 
 export type EventDays = Event[][];
+
+export type EventCategory = {
+  label: string,
+  color: string,
+  contrast?: boolean
```
RGBboy: I suggest making this required instead of optional and relying on undefined, then just add false to the categories that should not invert the color. Reason being this can now be three different values, `true` `false` and `undefined`. Makes it much clearer that it can only be two values `true` and `false`.

src/components/CategoriesPills.test.js
daralthus: This is a good question, it would look something like this:

android/app/src/main/java/com/pridelondonapp/MainApplication.java
```diff
@@ -32,6 +33,7 @@ public boolean getUseDeveloperSupport() {
             new LinearGradientPackage(),
             new SplashScreenReactPackage(),
             new MapsPackage(),
+            new LinearGradientPackage(),
```
frigus02: This is here twice.

android/settings.gradle
```diff
@@ -7,5 +7,7 @@ include ':react-native-maps'
 project(':react-native-maps').projectDir = new File(rootProject.projectDir, '../node_modules/react-native-maps/lib/android')
 include ':react-native-config'
 project(':react-native-config').projectDir = new File(rootProject.projectDir, '../node_modules/react-native-config/android')
+include ':react-native-linear-gradient'
+project(':react-native-linear-gradient').projectDir = new File(rootProject.projectDir, '../node_modules/react-native-linear-gradient/android')
```
frigus02: This is here twice as well.

ios/PrideLondonApp.xcodeproj/project.pbxproj
```diff
@@ -1064,6 +1095,20 @@
...
+			remoteRef = 5AB34E402077745700E0E793 /* PBXContainerItemProxy */;
+			sourceTree = BUILT_PRODUCTS_DIR;
+		};
+		5AB34E432077745700E0E793 /* libBVLinearGradient.a */ = {
+			isa = PBXReferenceProxy;
+			fileType = archive.ar;
+			path = libBVLinearGradient.a;
+			remoteRef = 5AB34E422077745700E0E793 /* PBXContainerItemProxy */;
+			sourceTree = BUILT_PRODUCTS_DIR;
+		};
```
frigus02: I suspect this might be here twice as well.

android/app/src/main/java/com/pridelondonapp/MainApplication.java
daralthus: Ah good catch!

declanslevin:
Tested on the following:
Galaxy S8 (8.0) Physical
iPhone 7+ (10.3.1) Physical
iPhone 6 (10.3.3) Physical
Galaxy S7 (6.0) BrowserStack
Pixel 2 (8.0) BrowserStack
iPhone X (11.0) BrowserStack


Found a few issues but overall looks good!

Issue #73, opened 2018-03-25
Date fields from Contentful is not validated in app


frigus02:
Contentful allows you to enter an endDate, which is before the startDate. This currently crashes the app.

How do we want to deal with this? Validate the dates in the app and swap them?

RGBboy:
Wip is being developed in #76

PR #106, opened 2018-04-08
Switch dates if end date is before start date

jonyardley:
As there is no validation in contentful if the `startTime` is after the `endTime` then we should switch them around as to not break elements in the UI.

I have also moved the locale into a separate module. (Probably should have done this in a separate PR but thought some housekeeping wouldn't go amiss).

Fixes: #73

## Pre-flight check-list
* [x] Documentation (N/A)
* [x] Unit tests

## Pre-merge check-list

* [ ] Tester approved

## TestPlan
- [ ] Create/change an event in contentful with the start date before the end date.
- [ ] App does not break and start and end date should be switched.

### DIFF
--- a/e2e/init.js
+++ b/e2e/init.js
@@ -2,7 +2,7 @@ import detox from "detox";
 import { detox as config } from "../package.json";
 
 // Set the default test timeout of 120s
-jest.setTimeout(120000);
+jest.setTimeout(300000);
 
 beforeAll(async () => {
   await detox.init(config);
--- a/src/data/get-asset-url.js
+++ b/src/data/get-asset-url.js
@@ -1,7 +1,7 @@
 // @flow
 import type { Asset, LocalizedFieldRef } from "./event";
 
-const locale = "en-GB";
+import locale from "../data/locale";
 
 export default (getAssetById: string => Asset) => (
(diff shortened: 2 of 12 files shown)


jonyardley:
@declanslevin I have moved this onto a new branch in favour of the old one to potentially fix a bad rebase. I have also increased the timeout for the ios-e2e which was failing.

jonyardley:
This is green now :)

declanslevin:
So this is working when I build the app locally (Android) by pulling the branch, but if I download the latest hockeyapp builds (alpha or beta) the dates aren't being corrected. It was working earlier (on both Android and iOS) when testing the hockeyapp builds, so it seems like maybe later commits are conflicting with the changes committed for this branch?

jonyardley:
hmm weird. Just done a merged from master. Maybe that will help?

declanslevin:
I think I just didn't understand completely how hockeyapp builds work... after some fiddling and help from Jan we figured it out.

Tested on:
(Physical)
Galaxy S8
iPhone 7+ 

(BrowserStack)
Galaxy S7

And it's working correctly (and the date filter crash is gone too)

Pending testing on some other iOS devices (once I can get the proper build from hockeyapp)

PR #122, opened 2018-04-13
Event Detail Initial Layout

RGBboy:
Part 1 of 14 [[UI] Events Listing - View Individual Event Details](https://trello.com/c/iW3T4PWb/128-ui-events-listing-view-individual-event-details)

This PR adds only the header from the design and the key image with the correct aspect ratio. The rest will come incrementally.

Along the way I also moved to use the same color strategy that is coming in #101. I.e. using the defined colors from Zeplin.

I suggest in the spirit of getting this all in incrementally we don't wait for test to review the individual PR's and instead get these in and do a review at the end.


src/components/ImageWithAspect.js
```diff
@@ -0,0 +1,42 @@
...
+class ImageWithAspect extends Component<Props, State> {
+  constructor() {
+    super();
+    this.state = {
+      height: 0
+    };
+  }
+  handleLayout = (event: { nativeEvent: { layout: { width: number } } }) => {
+    this.setState({
+      height: event.nativeEvent.layout.width / this.props.ratio
```
frigus02: There is a style prop called [`aspectRatio`](http://facebook.github.io/react-native/docs/layout-props.html#aspectratio). Did you try using this?

src/screens/EventDetailsScreen/IconButton.js
```diff
@@ -0,0 +1,24 @@
...
+import { Image, StyleSheet } from "react-native";
+import Touchable from "../../components/Touchable";
+
+type Props = {
+  onPress: Function,
+  source: Image.propTypes.source
+};
+
+const IconButton = ({ onPress, source }: Props) => (
+  <Touchable onPress={onPress}>
```
frigus02: This needs an `accessibilityLabel`. Otherwise the screen reader will not read any text for this button.

src/screens/EventDetailsScreen/Header.js
```diff
@@ -1,36 +1,26 @@
...
 };
 
-const Header = ({ onBackButtonPress, imageUrl }: Props) => (
-  <ImageBackground
-    style={styles.header}
-    source={{ uri: imageUrl }}
-    resizeMode="cover"
-  >
+const Header = ({ children }: Props) => (
+  <View style={styles.header}>
```
frigus02: On iOS you can tell the screen reader, that an element is a header. Maybe this should get `accessibilityTraits={["header"]}`.

See: http://facebook.github.io/react-native/docs/accessibility.html

src/components/ImageWithAspect.js
RGBboy: Amazing, I didn't even know this existed. I have updated using this instead.

PR #126, opened 2018-04-14
155 Support us page

frigus02:
[Trello - Support Pride Page](https://trello.com/c/7nGZf47s/155-support-pride-page)

image

It also addresses #133 for some screens.

## Test plan

* Go to the "Support us" tab.
* Do all kinds of weird shit on the page. It should not break.
* "Donate" and "Sponsor" buttons should open blank pages in the app.
* "Volunteer" and "Buy merchandise" buttons should open the browser on example.com.

## Pre-flight check-list

Before raising a pull request

* [x] Documentation
* [x] Unit tests

## Pre-merge check-list

* [x] Link to [Trello ticket](https://trello.com/c/7nGZf47s/155-support-pride-page)
* ~Any risky areas identified~
* [x] Test plan provided
* [ ] Tester approved

## Tester check-list

  * [x] Galaxy S7 (Android 6.0)
  * [ ] iPhone X (iOS 11.x)
  * [ ] iPhone 8 (11.x)
  * [ ] iPhone 7 (iOS 10.x)

  Optional:

  * [ ] Google Pixel 2
  * [ ] Galaxy S8 Plus
  * [ ] Galaxy S7 Edge
  * [ ] Galaxy S6
  * [x] iPhone 7+
  * [x] iPhone 6

* [x] Text-to-Voice (Android: Voice Assistant / iOS: Speak Selection)
* [x] Large Text (=<200%)

* [ ] Airplane mode
* [ ] 2G/3G Network Settings


frigus02:
This is still missing the URLs for "Volunteer" and "Buy merchandise" links. Also it is missing the shapes in the background of the 4 buttons.

But since this is the entry page to the Donate and Sponor pages, I think it makes sense to merge it like this and add the missing pieces later.

src/screens/SupportUsScreen/index.js
```diff
@@ -0,0 +1,115 @@
...
+      </View>
+    </ScrollView>
+  </View>
+);
+
+const styles = StyleSheet.create({
+  container: {
+    flex: 1
+  },
+  scrollContainer: {
```
RGBboy: The `backgroundColor` should be `whiteColor` if I am reading the design correctly. It looks slightly grey to me which I think is coming from the parent view.

src/screens/SupportUsScreen/SupportUsSectionHeader.js
```diff
@@ -0,0 +1,36 @@
...
+    aspectRatio: 2 / 1,
+    justifyContent: "flex-end",
+    alignItems: "flex-start"
+  },
+  title: {
+    backgroundColor: whiteColor,
+    color: lightNavyBlueColor,
+    margin: 16,
+    paddingHorizontal: 8,
+    lineHeight: 40
```
RGBboy: On iOS this looks a bit top heavy. I suggest using:

```
paddingTop: 4,
lineHeight: 36
```

Existing (with descender text):


With change (with descender text):

frigus02: Nice. Will try this. Text alignment seems to differ a bit between Android and iOS. So I will need to find a good middle gound, I think.

JunBadger:
- Is the background currently a grey? It looks off-white (and should be #fff)
- Is the left margin between the text and left hand side of CTA 20px? Looks slightly bigger in build
- Cell will upload the shapes for you @frigus02

frigus02:
Yeah, the background is gray. I will fix this and check the margin. It should be 20px, though.

frigus02:
* I fixed the background color.
* I aligned the text title text better (thanks @RGBboy, your suggestion works great)
* I fixed the padding of the CTA buttons.
* I added the background shapes.

There 2 remaining issues:
* I needed to give the shapes a general `borderRadius` instead of a radius only on the left/right side, because it's the only thing iOS supports.
* When the user resizes the font, it will increase the height of the CTA buttons. But the shapes don't resize with it.

Someone can pick this up during the week. Or I can have a look at this on friday.

cellyn-tan:
Downloaded the build re: background shapes on my phone (iPhone X) - looks like the shapes are not contained within the CT for  **Buy Merchandise** and **Partner with us**

img_7917

frigus02:
Weird. The reason is:
> When the user resizes the font, it will increase the height of the CTA buttons. But the shapes don't resize with it.

But I wonder what height the buttons have. Did you scale the font down (make it smaller)?

RGBboy:
@frigus02  What if we instead set an image as the background image? I would suggest that we just create larger images that look good when the button grows vertically. They don't need to scale, they just reveal more as the button grows.

frigus02:
Background image: Can work, I think. The donate button needs 2 background images, though. One for the left and one for the right side.

Larger images, which just reveal more as the button grows is a good idea. :+1:

RGBboy:
Apologies, I just saw that React Native does not support background images. I still think there is value in design providing bigger images if the button grows. I think you may be able to mitigate the overflow problem by setting the overflow on the Image to `hidden`

cellyn-tan:
@frigus02 @RGBboy Ah got it. The font size I set on my phone was slightly smaller than the default font size. How about if we set a minimum height for the CTAs?

frigus02:
React Native has basic support for background images using the `<ImageBackground>` component. We could maybe nest 2 of those:

```jsx
<ImageBackground source={bgLeft}>
  <ImageBackground source={bgRight}>
    Donate
  </ImageBackground>
</ImageBackground>
```

We need to see how we can align the images then. Might be worth a try.

frigus02:
I think the overflow solution sounds good for the images. I tried to just scale the images and it didn't look that bad, either 🙂

declanslevin:
There's a small typo for the Donate button:


It should say "Every donation keeps us marching"

PR #138, opened 2018-04-16
Re-using category labels

daralthus:
## Pre-flight check-list

Before raising a pull request

* [x] ~~Documentation~~
* [x] ~~Unit tests~~ (Already covered by snapshot tests)

## Pre-merge check-list

https://trello.com/c/iW3T4PWb/128-ui-events-listing-view-individual-event-details

## Notes

I would suggest a smaller font size for these actually as long categories like "Cabaret and Variety" wrap the next label to a new line:
image

### DIFF
--- a/src/components/CategoriesPills.js
+++ b/src/components/CategoriesPills.js
@@ -3,42 +3,21 @@ import React from "react";
 import { View, StyleSheet, ScrollView, PixelRatio } from "react-native";
 import LinearGradient from "react-native-linear-gradient";
 import type { StyleObj } from "react-native/Libraries/StyleSheet/StyleSheetTypes";
+import CategoryPill from "./CategoryPill";
 import Text from "./Text";
 import text from "../constants/text";
-import eventCategories from "../constants/event-categories";
-import locale from "../data/locale";
 
 import {
-  blackColor,
-  whiteColor,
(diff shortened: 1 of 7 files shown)


src/components/CategoriesPills.js
```diff
@@ -81,7 +60,11 @@ class CategoriesPills extends React.PureComponent<Props> {
               onContentSizeChange={this.scrollToLastPill}
             >
               {Array.from(selectedCategories).map(name => (
-                <CategoryPill key={name} name={name} />
+                <CategoryPill
+                  key={name}
+                  name={name}
+                  style={styles.categoryPill}
```
RGBboy: I'm hesitant to pass styles to a component. I have often found a better option is to wrap the component with layout components. This ties in with single responsibility, and also enables reuse of both this component and layout primitives. I'll let you decide for this PR, but I think we should move to creating layout components eventually.

daralthus: That's an interesting one. I think we agree that as long as it doesn't limit composability it won't be a problem. 
I might be biased towards keeping it like this, because I haven't seen a good example of layout components being extensively used. Not sure if this project will grow that much to have a design system where we can get to that granularity of re-use. 
I hope to come back to this when we have a few more use cases.

PR #143, opened 2018-04-17
Event details - event location

matt-paul:
## Pre-flight check-list

Before raising a pull request

* [x ] Documentation
* [x ] Unit tests

Note: We have updated the Contentful Event Schema with address fields in order to be able to display the address.  These are:

 - Address Line 1
 - Address Line 2
 - City
 - Postcode

All of these fields have been marked as optional within Contentful and within the event flow type.

## Pre-merge check-list

https://trello.com/c/iW3T4PWb


src/data/event.js
```diff
@@ -8,6 +8,10 @@ export type Event = {
     startTime: { [string]: string },
     endTime: { [string]: string },
     location: { [string]: { lat: number, lon: number } },
+    addressLine1: { [string]: string },
```
RGBboy: If these are optional, can we communicate that here? I think there may be a syntax to do so.

matt-paul:
@robbiemccorkell We have had to add fields to the Contentful content model to match the given Zeplin design.  Contentful currently only saves longitude and latitude. What do you think?

RGBboy:
With respect to the content model changes. Did you run this past Andy or Rob? I'm not sure if the content model is signed off, and changes may need to be communicated to the Pride PO.

src/data/event.js
matt-paul: 👍

matt-paul:
@RGBboy Chatted with Robbie and he is happy to add the fields in.   They are all now optional.

RGBboy:
Great. I would just update the flow types for the event to communicate that then I'll give this an approval.

PR #156, opened 2018-04-20
adding in numberOfLines prop to text component

### LINKED ISSUE #111: Entire name of events with very long titles are displayed on the home page, which distorts the layout of the events
Tested on multiple devices:
Galaxy S8, Galaxy Note Edge, iPhone 7+, iPhone 6, iPhone X (BrowserStack) 

The entire name of events with very long titles get displayed on the home page when the event is added to the Featured events list. This disrupts the layout of the featured events and they become misaligned, leaving large sections of whitespace:

trcywu:
## Pre-flight check-list
fixes #111 

Before raising a pull request

* [ ] Documentation
* [ ] Unit tests

## Pre-merge check-list

* [ ] Test plan provided
* [ ] Tester approved

## Tester check-list

  * [ ] Galaxy S7 (Android 6.0)
  * [ ] iPhone X (iOS 11.x)
  * [ ] iPhone 8 (11.x)
  * [ ] iPhone 7 (iOS 10.x)

  Optional:

  * [ ] Google Pixel 2
  * [ ] Galaxy S8 Plus
  * [ ] Galaxy S7 Edge
  * [ ] Galaxy S6
  * [ ] iPhone 7+
  * [ ] iPhone 6

* [ ] Large Text (=<200%)

* [ ] Airplane mode
* [ ] 2G/3G Network Settings


src/components/Text.js
```diff
@@ -38,6 +40,7 @@ const Text = ({
       style={[type && styles[type], style]}
       allowFontScaling={allowFontScaling}
       onLayout={onLayout}
+      numberOfLines={numberOfLines}
```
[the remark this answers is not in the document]
trcywu: yes even better!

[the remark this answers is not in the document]
robbiemccorkell: Good luck wrestling with flow! :P

src/components/Text.js
```diff
@@ -14,31 +14,21 @@ type Props = {
...
 
-const Text = ({
-  children,
-  type,
-  markdown,
-  style,
-  allowFontScaling,
-  onLayout
-}: Props) =>
+const Text = ({ children, type, markdown, style, ...otherProps }: Props) =>
```
RGBboy: you can remove children and style too, then refactor both Markdown and RnText below.

src/components/Text.js
```diff
@@ -14,31 +14,21 @@ type Props = {
   markdown?: boolean,
   style?: StyleObj,
   allowFontScaling?: boolean,
-  onLayout?: Function
+  onLayout?: Function,
+  numberOfLines?: number
```
robbiemccorkell: Now that extra props are encapsulated within `otherProps` do we still need these prop types?

Is it because flow complains that we're passing these props down and wants to know about them?

trcywu: yes, tried to remove them and flow complains

RGBboy: You might be able to do something by extending the RnText Prop type instead.

RGBboy: I think intersection type is what you want: https://flow.org/en/docs/types/intersections/

src/components/Text.js
```diff
@@ -14,31 +14,21 @@ type Props = {
...
-  allowFontScaling,
-  onLayout
-}: Props) =>
+const Text = ({ children, type, markdown, style, ...otherProps }: Props) =>
   markdown ? (
     <Markdown
       style={style}
       markdownStyles={{ ...textStyles, ...markdownStyles }}
-      onLayout={onLayout}
+      {...otherProps}
```
robbiemccorkell: We might get warnings passing all the props to the markdown element. From their source code it looks like they pass down any extra props to a `View` rather than `Text`.

I hadn't noticed earlier that this makes things a bit more tricky. Because some props like `onLayout` are accepted by `View`, but others like `allowFontScaling` won't be. If the dev build runs fine without warnings then this is fine I think. Otherwise we might have to be explicit just with this component like:

```
<Markdown
  style={style}
  markdownStyles={{ ...textStyles, ...markdownStyles }}
  onLayout={otherProps.onLayout}
>
```
etc

trcywu: debated this one with @frigus02 and wasn't sure so happy to revert back

frigus02: Yeah, we were not sure about this. It might be a bit confusing. But looks like flow is fine with it.

robbiemccorkell: I dunno really. If there aren't any warnings in the console, then I think it's all good as you've got it

src/components/Text.js
frigus02: I thought about this, too. Has the issue, that `Markdown` does not support all the same props. But it might still be fine. The only requirement is, that we can actually import the default text prop types. I noticed this they are not available for all components.

src/components/Text.js
robbiemccorkell: I remember at some point React added warnings when unknown props were passed to a component, but that might not be the case anymore.

src/components/Text.js
robbiemccorkell: 🎺

PR #164, opened 2018-04-21
Date picker accessibility

charypar:
<https://trello.com/c/PQIdwgZb/125-ui-events-listing-filter-events-list-by-date>

This should be the final piece of the date filtering story. It would really benefit from a test on an Android device. 

Works _reasonably_ well on iOS, although the navigation around the calendar is still pretty clunky. Honestly, I can't see visually impaired users use the filter too much 😞. 

Test with VoiceOver, ideally with Heading mode selected in the "rotor" - I used it to make the navigation around the main sections of the calendar a little easier (iOS calendar does similar things).

## Pre-flight check-list

Before raising a pull request

* [x] Unit tests

## Pre-merge check-list

* [x] Link to Trello ticket/GitHub issue (If applicable) - https://trello.com/c/PQIdwgZb/125-ui-events-listing-filter-events-list-by-date
* [x] Test plan provided
* [ ] Tester approved

## Tester check-list

  * [ ] Galaxy S7 (Android 6.0)
  * [ ] iPhone X (iOS 11.x)
  * [ ] iPhone 8 (11.x)
  * [ ] iPhone 7 (iOS 10.x)

  Optional:

  * [ ] Google Pixel 2
  * [ ] Galaxy S8 Plus
  * [ ] Galaxy S7 Edge
  * [ ] Galaxy S6
  * [ ] iPhone 7+
  * [ ] iPhone 6

* [ ] Large Text (=<200%)

* [ ] Airplane mode
* [ ] 2G/3G Network Settings

### DIFF
--- a/src/components/DateRangePickerDay.js
+++ b/src/components/DateRangePickerDay.js
@@ -2,6 +2,8 @@
 import React, { Component } from "react";
 import { View, TouchableWithoutFeedback, StyleSheet } from "react-native";
 import { equals } from "ramda";
+import { format } from "date-fns";
+
 import Text from "./Text";
 
 import {
@@ -98,11 +100,15 @@ export default class Day extends Component<DayProps> {
 
   render() {
     const { state, marking, date } = this.props;
+    const label = format(date.dateString, "dddd, MMMM D");
+    const traits = marking.selected ? ["button", "selected"] : ["button"];
 
     return (
       <TouchableWithoutFeedback
         onPress={this.onPress}
         onLongPress={this.onLongPress}
+        accessibilityTraits={traits}
+        accessibilityLabel={label}
       >
         <View style={styles.container}>
           {marking.selected && (
--- a/src/components/DateRangePickerDay.js
+++ b/src/components/DateRangePickerDay.js
@@ -109,6 +109,7 @@ export default class Day extends Component<DayProps> {
         onLongPress={this.onLongPress}
         accessibilityTraits={traits}
         accessibilityLabel={label}
+        accessibilityComponentType="button"
       >
         <View style={styles.container}>
           {marking.selected && (
--- a/src/constants/text.js
+++ b/src/constants/text.js
@@ -11,6 +11,18 @@ const showEvents = (numberOfEvents: number) => {
   return `Show ${numberOfEvents} events`;
 };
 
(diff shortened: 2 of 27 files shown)


charypar:
The main glaring problem with this is the back and forward buttons in the calendar, which have no labels, so they appear active, but don't explain what they do. I can raise a PR against the dependency to add those, but I'm not sure how quickly it will get accepted. Thoughts?

src/components/DateRangePickerDay.js
```diff
@@ -98,11 +100,15 @@ export default class Day extends Component<DayProps> {
...
     const { state, marking, date } = this.props;
+    const label = format(date.dateString, "dddd, MMMM D");
+    const traits = marking.selected ? ["button", "selected"] : ["button"];
 
     return (
       <TouchableWithoutFeedback
         onPress={this.onPress}
         onLongPress={this.onLongPress}
+        accessibilityTraits={traits}
+        accessibilityLabel={label}
```
frigus02: You also want to use `accessibilityComponentType="button"` for Android.

src/screens/EventsScreen/component.js
```diff
@@ -8,6 +8,7 @@ import FilterHeader from "../../components/ConnectedFilterHeader";
 import { bgColor } from "../../constants/colors";
 import { EVENT_DETAILS, EVENT_CATEGORIES_FILTER } from "../../constants/routes";
 import locale from "../../data/locale";
+import text from "../../constants/text";
```
robbiemccorkell: Unneeded import?

src/constants/text.js
```diff
@@ -11,6 +11,18 @@ const showEvents = (numberOfEvents: number) => {
   return `Show ${numberOfEvents} events`;
 };
 
+const showEventsLabel = (numberOfEvents: number) => {
```
frigus02: These labels are very similar to the normal ones. I wonder if we should speak to UX to get one unified text instead?

charypar: The reader one are quite a lot longer to be very descriptive, since you lack the visual context, I think the visual ones are fine like this... but we can have a chat

src/components/DateRangePickerDay.js
charypar: Done in c9bc188

PR #165, opened 2018-04-22
Upgrade React Native

markholland:
Would be good to keep up with the monthly React Native releases.

Installed `react-native-git-upgrade` npm package globally and then ran it.

[Changelog](https://github.com/redbadger/pride-london-app/compare/bump-rn?expand=1)

## Pre-merge check-list

* [ ] Test plan provided
* [ ] Tester approved

## Tester check-list

  * [ ] Galaxy S7 (Android 6.0)
  * [ ] iPhone X (iOS 11.x)
  * [ ] iPhone 8 (11.x)
  * [ ] iPhone 7 (iOS 10.x)

  Optional:

  * [ ] Google Pixel 2
  * [ ] Galaxy S8 Plus
  * [ ] Galaxy S7 Edge
  * [ ] Galaxy S6
  * [ ] iPhone 7+
  * [ ] iPhone 6

* [ ] Large Text (=<200%)

* [ ] Airplane mode
* [ ] 2G/3G Network Settings


markholland:
Will need someone to help out with flow errors

frigus02:
Seems like they improved their types for style sheets. If I understand correctly, we now need to import one of those:

```js
import type { ViewStyleProp } from "react-native/Libraries/StyleSheet/StyleSheet";
import type { TextStyleProp } from "react-native/Libraries/StyleSheet/StyleSheet";
import type { ImageStyleProp } from "react-native/Libraries/StyleSheet/StyleSheet";
import type { DangerouslyImpreciseStyleProp } from "react-native/Libraries/StyleSheet/StyleSheet";
```

We need to choose the correct one, based on the view where we will apply the styles to.

markholland:
That's what I figured but when try any of them I get a cryptic error

`Cannot create SupportUsButton element because inexact ____StyleSheetInternalStyleIdentifier_Internal [1] is incompatible
with exact LayoutStyle [2] in property style.`

src/components/Text.js
```diff
@@ -1,39 +1,45 @@
...
 export type TextType = "h1" | "h2" | "h3" | "h4" | "text" | "small" | "price";
+export type TextColor = "lightNavyBlueColor";
 
 type Props = {
-  type?: TextType,
-  color?: "lightNavyBlueColor",
-  markdown?: boolean,
+  type: TextType,
+  color: ?TextColor,
+  markdown: boolean,
```
robbiemccorkell: Is this not an optional prop anymore? We've still got a default for it below.

Or has the way we define these changed in the new flow?

frigus02: Well, this is a flow special. I definitely need to find the docs for this:

Normally you define an optional property with the question mark (`markdown?: boolean`). When used in a react component, flow then requires you to define a default prop for this. When you use the prop in the component it will always be defined, since the default value is used. However flow still thinks the value can be undefined.

Interestingly, only for component classes, there is another way. As soon as you define a default prop for any prop, it automatically becomes optional for flow. It does not need to be marked as optional using the question mark. Then flow knows that the prop will not be undefined when you use it. Only in this case the default value for `color` is `undefined`. That's why I put the question mark in front of the type.

As I said, I did not read this in any documentation. I actually found this by accident. So this might not be entirely correct.

RGBboy:
The main things that changed are the move of the TextLink component into the EventDetailsScreen folder and the addition of the blackColor as the default color for our Text component.

RGBboy:
Note: When running in dev mode, a warning is presented:

```
Warning: isMounted(...) is deprecated in plain JavaScript React classes...
```

I don't think we are using this directly, so I am guessing one of our dependencies is. I'll have a look thins morning.

frigus02:
@RGBboy Does this mean we can merge?

RGBboy:
We can, however, in dev we will get the yellow banner unless we specifically mute that warning (as described above)

frigus02:
Ah. Got it. I would say, lets disable the yellow warning. Then merge this PR.

PR #166, opened 2018-04-22
Follow latest CircleCI docs for pre running simulator

markholland:
CircleCI finally added [2.0 documentation](https://circleci.com/docs/2.0/testing-ios/#pre-starting-the-simulator
) for pre launching the Simulator.

Set out to try and get Detox running stable. Not much success but made some minor build time reductions on OSX machines.

## Pre-merge check-list

* [ ] Test plan provided
* [ ] Tester approved

## Tester check-list

  * [ ] Galaxy S7 (Android 6.0)
  * [ ] iPhone X (iOS 11.x)
  * [ ] iPhone 8 (11.x)
  * [ ] iPhone 7 (iOS 10.x)

  Optional:

  * [ ] Google Pixel 2
  * [ ] Galaxy S8 Plus
  * [ ] Galaxy S7 Edge
  * [ ] Galaxy S6
  * [ ] iPhone 7+
  * [ ] iPhone 6

* [ ] Large Text (=<200%)

* [ ] Airplane mode
* [ ] 2G/3G Network Settings


markholland:
Reluctant to put this to review because now e2e seems to fails every time

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "pride-london-app",
-  "version": "1.0.5",
+  "version": "1.0.6",
```
markholland: Quickest way to get around iTunes Connect saying a build exists when it doesn't

markholland:
Unfortunately this doesn't really help on e2e but CI caching improvements should be good. Also chance to bump version as ITunes Connect thinks that build 91 of 1.0.5 already exists when it clearly doesn't.

.circleci/config.yml
```diff
@@ -234,13 +262,31 @@ jobs:
...
+              npm i -g react-native-cli
+      - restore_cache:
+          key: osx-dependencies-{{ checksum "package.json" }}
+      - run:
+          name: Install dependencies
+          command: npm install
+      - save_cache:
+          paths:
+            - /Users/distiller/project/node_modules
+          key: osx-dependencies-{{ checksum "package.json" }}
```
frigus02: It looks like you use the same cache for `get-cache-osx` and `save-cache-osx` steps (except the path to package.json). Is this intended?

markholland: The working directory is different for e2e as it’s run in the project root. I couldn’t see a way to pass the key and they need to coincide otherwise it would be a different cache for e2e and build jobs. Which isn’t too bad but because only e2e uses this key then there wasn’t much point in pulling out to an alias so I’ve left it inside the job

PR #173, opened 2018-04-25
Spike: Try out Fabric Answers (Analytics)

RGBboy:
Spike to try get Fabric Answers working in our Alpha builds.


RGBboy:
Seems that iOS is not working. Android is though. I have tried a couple of things so far. I did also find this: https://stackoverflow.com/questions/47325958/ios-fabric-crashlytics-answers-kit-not-collecting-data where it is suggested to remove the apps from Fabric and re-add them. Not sure of the impact on API keys etc.

markholland:
Keys are per organisation so not tied to apps. But adding apps to Fabric is an annoying process

RGBboy:
Closing in favor of #242

PR #177, opened 2018-04-25
Favourite Event - animated button

RGBboy:
First part of [Trello 53 - Favourite Events](https://trello.com/c/EGwManv3/53-events-listing-flagging-favourite-events). This PR adds Lottie and a very simple animated button to the Event Details screen. Further functionality will be added in later PRs.

I have tested locally on iOS and Android. I did face an issue with iOS which now seems to be resolved. However I would like another dev to pull this and try run `yarn run-ios` and report back any problems.

It would also be good to check more devices so that we can confirm this lib is going to give us what we need.

## Tester check-list

  * [x] Galaxy S7 (Android 6.0)
  * [x] iPhone X (iOS 11.x)
  * [x] iPhone 8 (11.x)
  * [x] iPhone 7 (iOS 10.x)

Optional:

* [x] Google Pixel 2
* [x] Galaxy S8 Plus
* [ ] Galaxy S7 Edge
* [x] Galaxy S6
* [x] iPhone 7+
* [x] iPhone 6


declanslevin:
Tested on physical Galaxy S8 and iPhone 7+/6, the rest were BrowserStack devices and the animation is working correctly on everything tested

android/app/build.gradle
```diff
@@ -222,6 +222,7 @@ repositories {
 }
 
 dependencies {
+    compile project(':lottie-react-native')
```
frigus02: Can you change this `compile` to `implementation`? Unfortunately react native still generates the older gradle syntax (`compile`), so we need to manually change this.

src/screens/EventDetailsScreen/component.js
```diff
@@ -20,7 +21,6 @@ import text from "../../constants/text";
 import type { Event, LocalizedFieldRef } from "../../data/event";
 import locale from "../../data/locale";
 import chevronLeftWhite from "../../../assets/images/chevron-left-white.png";
-import heartWhite from "../../../assets/images/heart-white.png";
```
frigus02: Can we delete this icon or do we still use it somewhere else?

frigus02:
This looks amazing. But it feels like it takes a while until the animation starts.

src/screens/EventDetailsScreen/component.js
RGBboy: It can go. Nice catch.

RGBboy:
Yeah, I think there will need to be some tweaks. This PR was more proof that we could get this working across platforms. I will add a task to tweak the animations once this is all in.

android/app/build.gradle
RGBboy: done

PR #180, opened 2018-04-27
Favourite Event - reducer

RGBboy:
Second part of [Trello 53 - Favourite Events](https://trello.com/c/EGwManv3/53-events-listing-flagging-favourite-events). This PR adds the Saved Events reducer and hooks it up with the event details screen. Currently this just persists the state in memory. A future PR will commit this to local storage too.

**Note:** It is known that the animation feels delayed and that when looking at an even card that displays a saved event, occassionally there is a delay in showing the saved state in the heart. These will be addressed in a future PR.

## Tester check-list

  * [ ] Galaxy S7 (Android 6.0)
  * [ ] iPhone X (iOS 11.x)
  * [ ] iPhone 8 (11.x)
  * [ ] iPhone 7 (iOS 10.x)

Optional:

* [ ] Google Pixel 2
* [ ] Galaxy S8 Plus
* [ ] Galaxy S7 Edge
* [ ] Galaxy S6
* [ ] iPhone 7+
* [ ] iPhone 6


src/actions/saved-events.js
```diff
@@ -0,0 +1,15 @@
+// @flow
+import type { Dispatch } from "redux";
+
+type AddSavedEventAction = { type: "ADD_SAVED_EVENT", id: string };
+type RemoveSavedEventAction = { type: "REMOVE_SAVED_EVENT", id: string };
+
+export type SaveEventAction = AddSavedEventAction | RemoveSavedEventAction;
```
frigus02: Should this be named `SavedEventAction` (past tense) for consistency?

src/components/SaveEventButton.js
```diff
@@ -0,0 +1,81 @@
...
+    onDark: false
+  };
+
+  state = {
+    progress: new Animated.Value(0)
+  };
+
+  componentDidMount() {
+    if (this.props.active) {
+      this.state.progress = new Animated.Value(1);
```
frigus02: I'm confused. Should we be using `setState({ progress: new Animated.Value(1) });` here?

src/components/SaveEventButton.js
```diff
@@ -0,0 +1,81 @@
...
+        useNativeDriver: true
+      }).start();
+      return;
+    }
+    Animated.timing(this.state.progress, {
+      toValue: value,
+      duration: 0,
+      easing: Easing.linear,
+      useNativeDriver: true
+    }).start();
```
frigus02: Why is the animation with duration `0` needed? The start value for `state.progress` should always be correct because of `componentDidMount`. So all changes to `active` should be covered already. And when another prop than `active` changes while an animation is in progress, this will most likely interrupt the animation.

src/components/SaveEventButton.js
```diff
@@ -0,0 +1,81 @@
...
+
+  handlePress = () => {
+    this.props.onPress(!this.props.active);
+  };
+
+  render() {
+    const source = this.props.onDark ? heartAnimationLight : heartAnimationDark;
+    return (
+      <Touchable
+        accessibilityLabel="Favourite"
```
frigus02: I wonder if the `accessibilityLabel` needs to include the current state of the button, i.e. "Add to favorites"/"Remove from favorites".

src/actions/saved-events.js
frigus02: I just noticed, that we use a `StandardAction` type in the other actions. Does it make sense to use it here as well? I don't have strong feelings about this, though. Just nitpicking 🙂

src/components/SaveEventButton.js
RGBboy: This is to toggle the button to a value state immediately. For example when the component mounts we do not want the animation to play. This should only happen when the button is not active and changes to active.

src/actions/saved-events.js
frigus02: Cool. Makes sense to me.

src/components/SaveEventButton.js
frigus02: But isn't this what `componentDidMount` is for?

RGBboy: I have refactored, but this still needs to be present for the toggling active from true to false case.

src/components/SaveEventButton.js
RGBboy: We warn against that. This was a bit of a hack, I have now updated to use `getDerivedStateFromProps`.

src/components/SaveEventButton.js
RGBboy: I will ask. I will defer for now, and add a note to trello so this is considered.

src/components/SaveEventButton.js
```diff
@@ -0,0 +1,87 @@
...
+        }).start();
+      } else if (this.props.active !== prevProps.active) {
+        Animated.timing(this.state.progress, {
+          toValue: value,
+          duration: 0,
+          easing: Easing.linear,
+          useNativeDriver: true
+        }).start();
+      }
+    }
```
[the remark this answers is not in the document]
RGBboy: Added a comment, also refactored to make the code a bit more concise.

PR #186, opened 2018-04-28
82 Donate page

frigus02:
[Trello Ticket - Donate](https://trello.com/c/XmsDjp8f/82-donate)

## Pre-flight check-list

Before raising a pull request

* [ ] Documentation
* [ ] Unit tests

## Pre-merge check-list

* [ ] Test plan provided
* [ ] Tester approved

## Tester check-list

  * [ ] Galaxy S7 (Android 6.0)
  * [ ] iPhone X (iOS 11.x)
  * [ ] iPhone 8 (11.x)
  * [ ] iPhone 7 (iOS 10.x)

  Optional:

  * [ ] Google Pixel 2
  * [ ] Galaxy S8 Plus
  * [ ] Galaxy S7 Edge
  * [ ] Galaxy S6
  * [ ] iPhone 7+
  * [ ] iPhone 6

* [ ] Large Text (=<200%)

* [ ] Airplane mode
* [ ] 2G/3G Network Settings


frigus02:
Unfortunately accessibility for the text input field in TalkBack is really bad.

A native input field on Android (tested in the default Contacts app):
* Announces it's label and value when it receives screen reader focus.
* Double tap activates it: shows caret and brings up keyboard.
* When activated and the screen reader focus is moved away, it stays activated. Double tap on any other activatable element deactivates the input and activates the new element.

The react native TextInput however:
* Announces the `accessibilityLabel` only when the value is empty and there is no placeholder text. When either a value or placeholder is visible, only this is announced.
* Double tap does not activate it, even though this is what the screen reader tells you to do. You need to double tab and hold to activate it. Then it shows the caret and brings up the keyboard. (Bug: https://github.com/facebook/react-native/issues/17624)
* When activated and the screen reader focus is moved away, it stays activated. Double tap on any other activatable element only deactivates the input. You need to double tap again to activate the new element.

It would be great if anyone could test this with VoiceOver as I don't have an iPhone.

frigus02:
@cellyn-tan Yeah, I will have another look at this. It was surprisingly hard to get it working on iOS and Android. But there has to be a way :-)

src/constants/colors.js
```diff
@@ -32,6 +32,8 @@ export const headerBgColor = "#A5A5A5";
 export const transparent = "transparent";
 
 export const textColor = "#272727";
+export const inputFieldBorderColor = "rgb(204, 204, 204)";
+export const inputFieldPlaceholderColor = "rgb(204, 204, 204)";
```
RGBboy: Can we avoid this abstraction and instead name as the color from Zeplin? If a named color does not exist, chat with design and get them to add it to Zeplin. Ideally when creating our components we should just be following the design, and using the same language. I.e. the inputField border is "lightGrey", so in our inputField component, we see `borderColor: lightGreyColor`.

There is a couple of nice side effects of doing this:
* We can do a find across the project and inform design as to all the components that use a given color, enabling them to make more informed decisions if these need to change in future.
* Design can actually check the code for our components and ensure the correct colors are used.

src/screens/DonateScreen/SegmentedControl.js
```diff
@@ -0,0 +1,87 @@
...
+import Text from "../../components/Text";
+import Touchable from "../../components/Touchable";
+import {
+  inputFieldBorderColor,
+  eucalyptusGreenColor
+} from "../../constants/colors";
+
+type Props = {|
+  onSelectedIndexChange: number => void,
+  selectedIndex: ?number,
```
RGBboy: Can we follow the same API as the `Picker` component and use `selectedValue` and `onValueChange` instead? We could also add a SegmentItem component to follow the same child based API too, especially if we need to add accessibility labels for each item.

RGBboy:
There is also this issue opened that this screen is affected by: https://github.com/redbadger/pride-london-app/issues/188. Essentially the CTA on the iPhone X is too close to the bottom of the screen. This could still be merged without that fix and we can address all of the places this appears in one PR.

src/constants/colors.js
frigus02: This wasn't a named color in Zeplin, yet. I will check with design which one to use or if we create a new one.

src/screens/DonateScreen/SegmentedControl.js
RGBboy: Personally I prefer the strategy of defining values/options in JSX. Especially when you restrict the children type to a particular class. I'll let you make the call.

src/constants/colors.js
RGBboy: TY

frigus02: @JunBadger Will add a new medium gray color, which we can use for disabled buttons, the input field border and placeholder color.

src/screens/DonateScreen/SegmentedControl.js
frigus02: I changed the API to look the same as the `Picker` and I actually quite like it.

frigus02:
@cellyn-tan The page scrolls down to reveal the Donate button now when the user focuses the input field. Turned out to be very easy.

Only thing missing now is accessibility.

frigus02:
On iOS it seems fine and Android is just bugged. I suggest we merge like this and do another PR to improve accessibility.

PR #192, opened 2018-04-30
Saved events screen

zoeabryant:
### In this PR
- Saved events icon goes to an event listing type screen
- Saved events page only shows saved events

### Future PRs
- No saved events screen
- Need to confirm what is meant to happen when you select an event, then press the back button. which screen should you see?

Before raising a pull request

* [?] Documentation
* [x] Unit tests

## Pre-merge check-list

* [x] Link to Trello ticket/GitHub issue (If applicable) https://trello.com/c/CPa5xWwS
* [ ] Test plan provided
* [ ] Tester approved

## Tester check-list

  * [ ] Galaxy S7 (Android 6.0)
  * [ ] iPhone X (iOS 11.x)
  * [ ] iPhone 8 (11.x)
  * [ ] iPhone 7 (iOS 10.x)

  Optional:

  * [ ] Google Pixel 2
  * [ ] Galaxy S8 Plus
  * [ ] Galaxy S7 Edge
  * [ ] Galaxy S6
  * [ ] iPhone 7+
  * [ ] iPhone 6

* [ ] Large Text (=<200%)

* [ ] Airplane mode
* [ ] 2G/3G Network Settings

### DIFF
--- a/src/screens/SavedEventListScreen/component.js
+++ b/src/screens/SavedEventListScreen/component.js
@@ -6,6 +6,7 @@ import type { SavedEvents, EventDays } from "../../data/event";
 import type { LocalizedFieldRef } from "../../data/localized-field-ref";
 import EventList from "../../components/EventList";
 import Header from "../../components/Header";
+import text from "../../constants/text";
 import ContentPadding from "../../components/ContentPadding";
 import IconButton from "../../components/IconButton";
 import Text from "../../components/Text";
@@ -64,7 +65,7 @@ class SavedEventListScreen extends PureComponent<Props> {
         <Header backgroundColor={lightNavyBlueColor}>
           <ContentPadding style={styles.headerContent}>
             <Text type="h2" style={styles.headerTitle}>
-              Saveeeeed events
+              {text.savedEventsTitle}
             </Text>
           </ContentPadding>
         </Header>
--- a/src/screens/SavedEventListScreen/component.js
+++ b/src/screens/SavedEventListScreen/component.js
@@ -1,0 +1,111 @@
+// @flow
+import React, { PureComponent } from "react";
+import { StyleSheet, View } from "react-native";
+import type { NavigationScreenProp, NavigationState } from "react-navigation";
+import type { SavedEvents, EventDays } from "../../data/event";
+import type { LocalizedFieldRef } from "../../data/localized-field-ref";
+import EventList from "../../components/EventList";
+import Header from "../../components/Header";
+import ContentPadding from "../../components/ContentPadding";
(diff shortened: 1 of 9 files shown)


src/components/StandaloneHeader.js
```diff
@@ -0,0 +1,35 @@
...
+import Header from "./Header";
+import ContentPadding from "./ContentPadding";
+import Text from "./Text";
+import { whiteColor, lightNavyBlueColor } from "../constants/colors";
+
+export type Props = {
+  title: string
+};
+
+const StandaloneHeader = ({ title }: Props) => (
```
RGBboy: I suggest just inlining this into the SavedEventsScreen for now. Once this screen is in master we can then look at all of the headers and make the appropriate abstraction.

src/screens/SavedEventListScreen/component.js
```diff
@@ -0,0 +1,77 @@
...
+        ) : (
+          <EventList
+            locale={locale}
+            events={events}
+            savedEvents={savedEvents}
+            addSavedEvent={addSavedEvent}
+            removeSavedEvent={removeSavedEvent}
+            refreshing={refreshing}
+            onRefresh={() => {
+              updateEvents();
```
RGBboy: You might be able to simplify this by just passing `updateEvents` instead of wrapping in an arrow. I.e.

```
onRefresh={updateEvents}
```

zoeabryant:
hey @declanslevin yup, totally agree! I will approach it as part of a future PR

src/components/StandaloneHeader.js
RGBboy: Yeah, not that seriously for the duplication problem. I still think we should not abstract just yet as this kind of header appears in about 5 places currently. I think we should just address that all in one PR to make it clear as to the approach.

RGBboy:
@declanslevin @zoeabryant with respect to navigating to the incorrect screen I suggest we create an issue for it so we can decide on a good strategy here. I imagine we suffer from this in a number of places. I.e. home, featured events, saved events and event listings.

zoeabryant:
@declanslevin @RGBboy it's actually part of the A/C for this ticket:

> - When you tap on an event in your favourites list, you'll go to the Individual Event Details. Going back from the Individual Event Details page should take you back to the list of favourite events screen. 
so i think it's a good one for me to address! Not sure how to go about it yet, I am planning to look at it in a bit :)

src/components/StandaloneHeader.js
zoeabryant: I've reverted this now :)

PR #194, opened 2018-05-01
iTunes Connect query without app version is too vague

markholland:
We started getting an error when trying to retrieve the current build version in iTunes Connect.

https://circleci.com/gh/redbadger/pride-london-app/3571

We use this to know what a valid build number would be before starting a release build. This has worked fine up until now but a recent iTunes issue caused build `91` of `1.0.5` to not be accepted. To get around that initial issue we increased our app version to `1.0.6` and build `91` of this version was submitted succesfully.

But as of today iTunes now shows the failed `91` build of `1.0.5` so when we query for the latest build number there are two versions with build number `91` leading to iTunes not knowing which version we mean.

`You have to specify a new version number, as there are multiple to choose from`

The fix is to start passing the app version to this call, we get the current version from the Xcode project.

### DIFF
--- a/android/app/build.gradle
+++ b/android/app/build.gradle
@@ -123,7 +123,7 @@ android {
         minSdkVersion versions.minSdk
         targetSdkVersion versions.targetSdk
         versionCode androidVersionCode
-        versionName "1.0.6"
+        versionName "1.0.7"
         ndk {
             abiFilters "armeabi-v7a", "x86"
         }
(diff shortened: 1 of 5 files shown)


markholland:
Agh, ran this locally and the build numbers now correctly reset to 1 each version but we already have some builds for 1.0.6 I'm going to push a version bump and we should be good moving forwards.

PR #198, opened 2018-05-01
Update favourite animation with shorter lead in

RGBboy:
[Favourite Events](https://trello.com/c/EGwManv3/53-events-listing-flagging-favourite-events): Minor design tweaks to lessen the delay of the favourite animation on both the event details screen and the event listings screens.


declanslevin:
screenshot_20180501-153035_pride alpha
There is a tiny bit of the previous colour in the rainbow left in the top left of the favourite icon - as if the animation pauses just before it is finished.

It is barely visible on the Event details page but it is a lot easier to see on the events listing page

Tested on Galaxy S8 and iPhone 7+

RGBboy:
I have just pushed a change to add one more frame to the animation in the hopes it removes this artefact.

frigus02:
I just tested it on my phone. The vibration feels very strange. I think this is because no other app I know has a vibration for a normal touch. Vibration is usually used for a long press to indicate that you pressed long enough. Stuff like favorites usually trigger a toast message when pressed. I can only speak for Android, though. Maybe this is good for iOS.

RGBboy:
Yeah, we can always decrease the vibration amount, or turn it off on Android if appropriate. I'll leave it for now, and we can make judgement in UXD review.

PR #201, opened 2018-05-01
Saved events: no events screen, accessibility labels for the 💚 button

zoeabryant:
## Pre-flight check-list

depends on https://github.com/redbadger/pride-london-app/pull/192

Before raising a pull request

* [x] Documentation
* [x] Unit tests

## Pre-merge check-list

* [x] Link to Trello ticket/GitHub issue (If applicable)  https://trello.com/c/CPa5xWwS
* [ ] Test plan provided
* [ ] Tester approved

## Tester check-list

  * [ ] Galaxy S7 (Android 6.0)
  * [ ] iPhone X (iOS 11.x)
  * [ ] iPhone 8 (11.x)
  * [ ] iPhone 7 (iOS 10.x)

  Optional:

  * [ ] Google Pixel 2
  * [ ] Galaxy S8 Plus
  * [ ] Galaxy S7 Edge
  * [ ] Galaxy S6
  * [ ] iPhone 7+
  * [ ] iPhone 6

* [ ] Large Text (=<200%)

* [ ] Airplane mode
* [ ] 2G/3G Network Settings


src/screens/SavedEventListScreen/component.test.js
```diff
@@ -1,10 +1,10 @@
-// @flow
```
zoeabryant: If I keep flow in, it complains about my fake events. I tried copying the events from EventsList.test.js but complained about that as well. Ultimately decided it was adding value in this particular file.

src/constants/text.js
```diff
@@ -107,5 +107,9 @@ export default {
   zeroSelected: "0 selected",
   showEvents,
   clearAll: "Clear all",
-  filterEvents: "Filter Events"
+  filterEvents: "Filter Events",
+  noSavedEventsTitle: "Where is the love?",
+  noSavedEventsPart1: "No favourites yet? Tap the",
+  noSavedEventsPart2: "on any events you like to save them here.",
+  noSavedEventsButton: "Find Events"
```
RGBboy: To make this more discoverable, I suggest starting these with the screen name (`savedEvents...`) for example `savedEventsEmptyTitle`.

src/screens/SavedEventListScreen/component.test.js
RGBboy: I suggest leaving this on. Otherwise, we face the issue in future when the data model gets updated that our tests are not testing with the correct data structures.

src/screens/SavedEventListScreen/__snapshots__/NoSavedEvents.test.js.snap
```diff
@@ -0,0 +1,77 @@
...
+          "paddingBottom": 24,
+          "textAlign": "center",
+        }
+      }
+      type="text"
+    >
+      No favourites yet? Tap the
+       
+      <Image
+        accessibilityLabel="'save this event' button"
```
RGBboy: Should this label be put in the text constants?

src/screens/SavedEventListScreen/NoSavedEvents.js
```diff
@@ -0,0 +1,80 @@
...
+import strings from "../../constants/strings";
+
+type Props = {
+  navigation: NavigationScreenProp<NavigationState>
+};
+
+class NoSavedEvents extends PureComponent<Props> {
+  eventList = () => {
+    this.props.navigation.navigate(EVENT_LIST, {
+      title: strings.featuredEventsTitle
```
RGBboy: I don't think the title is needed? Also `featuredEventsTitle` is not the correct title for the event list screen 😉

src/screens/SavedEventListScreen/NoSavedEvents.js
```diff
@@ -0,0 +1,80 @@
...
+import { Image, StyleSheet, View } from "react-native";
+import type { NavigationScreenProp, NavigationState } from "react-navigation";
+import Text from "../../components/Text";
+import ContentPadding from "../../components/ContentPadding";
+import ButtonPrimary from "../../components/ButtonPrimary";
+import whiteCheck from "../../../assets/images/whiteCheck.png";
+import iconSave from "../../../assets/images/save.png";
+import noSavedEvents from "../../../assets/images/howToSaveEvents.png";
+import text from "../../constants/text";
+import { cardBgColor, titleTextColor } from "../../constants/colors";
```
RGBboy: We have slowly been migrating away from using abstracted colors and toward using the named colors from Zeplin. Can you do this here too?

src/screens/SavedEventListScreen/NoSavedEvents.js
```diff
@@ -0,0 +1,80 @@
+// @flow
+import React, { PureComponent } from "react";
+import { Image, StyleSheet, View } from "react-native";
+import type { NavigationScreenProp, NavigationState } from "react-navigation";
+import Text from "../../components/Text";
+import ContentPadding from "../../components/ContentPadding";
+import ButtonPrimary from "../../components/ButtonPrimary";
+import whiteCheck from "../../../assets/images/whiteCheck.png";
```
RGBboy: This is unused.

src/screens/SavedEventListScreen/NoSavedEvents.js
```diff
@@ -0,0 +1,80 @@
...
+    flexWrap: "wrap",
+    marginHorizontal: 12
+  },
+  image: {
+    alignSelf: "center"
+  },
+  title: {
+    color: titleTextColor,
+    textAlign: "center",
+    paddingBottom: 12
```
RGBboy: We have a column component that can do this kind of spacing for you. Might be worth checking out to avoid many paddings.

src/screens/SavedEventListScreen/component.test.js
zoeabryant: the linter is complaining about it too :D I'll have another look :)

src/screens/SavedEventListScreen/NoSavedEvents.js
zoeabryant: thanks, I'll have a look!

zoeabryant: yes, that made things easier, thanks @RGBboy !

src/screens/SavedEventListScreen/component.test.js
zoeabryant: .... I've just realised `EventList.test.js` also doesn't have flow in it, that'd be why it's failing the flow test!

src/constants/text.js
```diff
@@ -107,5 +107,12 @@ export default {
...
   clearAll: "Clear all",
-  filterEvents: "Filter Events"
+  filterEvents: "Filter Events",
+  noSavedEventsTitle: "Where is the love?",
+  noSavedEventsPart1: "No favourites yet? Tap the",
+  noSavedEventsSaveButtonAlt: "'save this event' button",
+  noSavedEventsPart2: "on any events you like to save them here.",
+  noSavedEventsButton: "Find Events",
+  unSaveEvent: "remove event from saved list",
+  saveEvent: "save this event"
```
RGBboy: I think you missed my comment about discoverability of these.

> To make this more discoverable, I suggest starting these with the screen name (`savedEvents...`) for example `savedEventsEmptyTitle`.

I would apply the same to the `unSaveEvent ` and `saveEvent`

src/components/SaveEventButton.test.js
```diff
@@ -3,13 +3,31 @@ import React from "react";
...
 it("renders correctly", () => {
   const output = shallow(<SaveEventButton active={false} onPress={() => {}} />);
   output.setState({ progress: new Animated.Value(0) });
   expect(output).toMatchSnapshot();
 });
 
+it("has correct accessibilityLabel when active=false", () => {
+  const output = shallow(<SaveEventButton active={false} onPress={() => {}} />);
+  output.setState({ progress: new Animated.Value(0) });
+  expect(output.find(Touchable).props().accessibilityLabel).toEqual(
```
RGBboy: I would use a snapshot test instead. You can simplify these 3 tests (including  `it("renders correctly", () => {` in to just two. One for active, one for inactive.

src/screens/SavedEventListScreen/NoSavedEvents.js
```diff
@@ -0,0 +1,66 @@
...
+import React, { PureComponent } from "react";
+import { Image, StyleSheet, View } from "react-native";
+import type { NavigationScreenProp, NavigationState } from "react-navigation";
+import Text from "../../components/Text";
+import ButtonPrimary from "../../components/ButtonPrimary";
+import LayoutColumn from "../../components/LayoutColumn";
+import iconSave from "../../../assets/images/save.png";
+import noSavedEvents from "../../../assets/images/howToSaveEvents.png";
+import text from "../../constants/text";
+import { titleTextColor } from "../../constants/colors";
```
RGBboy: I think you missed the comment on color too.

> We have slowly been migrating away from using abstracted colors and toward using the named colors from Zeplin. Can you do this here too?

src/screens/SavedEventListScreen/component.test.js
```diff
@@ -5,17 +5,88 @@ import { shallow } from "enzyme";
 import Component from "./component";
 import EventList from "../../components/EventList";
 import Loading from "../../components/Loading";
+import NoSavedEvents from "./NoSavedEvents";
 
 const navigation: NavigationScreenProp<NavigationState> = ({
   navigate: () => {}
 }: any);
 
 describe("EventsScreen Component", () => {
```
RGBboy: Can you change the name of this describe to the name of the screen please?

src/constants/text.js
zoeabryant: `unSaveEvent` and `saveEvent` are used on the 💚button on all Event tiles - on the EventDetails, EventListing etc, so they don't really exclusively belong to one screen.

Could change it to `SaveEventButtonSaveEvent` `SaveEventButtonUnSaveEvent`?

RGBboy: Yeah, the component name makes sense.

src/components/SaveEventButton.test.js
RGBboy: This is kind of the point of snapshot tests. What is useful is when others make changes in future. You can see when they update the snapshot the assertion has changed. Leaving both statements in means the test needs to be updated in two places if the copy changes.

zoeabryant: ok, both are just snapshot tests now. I've seen both approaches to snapshot tests - just snapshot tests, snapshot tests alongside specific property tests, they both have pros and cons. Not too worried about it for this test :)

PR #208, opened 2018-05-02
E2E 🔥

markholland:
Added some stuff to help diagnose issues with e2e.

- Seen multiple accounts of Yarn not playing well with Detox so have moved the e2e job to use npm entirely with it's own cache.
- Added screenshot taking after each test, these are stored in CI artifacts.
- E2E no longer requires the build job so will start straight away.
- Fixed some glaring issues made visible by recent CI config clean up 😄


.circleci/config.yml
```diff
@@ -154,7 +173,7 @@ set-node-version-macos: &set-node-version-macos
     name: Set node version
     command: |
       yarn global add n
-      sudo n 9.
+      sudo n 9.3
```
markholland: 🤦‍♂️

package.json
```diff
@@ -10,9 +10,10 @@
       "node node_modules/react-native/local-cli/cli.js run-android --appId org.prideinlondon.festival.dev --variant devDebug",
     "test": "jest src --maxWorkers=2",
     "test:e2e": "yarn test:e2e:build && yarn test:e2e:run",
-    "test:e2e:dev": "detox test --reuse",
+    "test:e2e:dev":
+      "detox test --reuse --loglevel verbose --debug-synchronization 1000",
     "test:e2e:run":
-      "detox test -c ios.sim.debug --loglevel verbose --debug-synchronization",
+      "detox test -c ios.sim.debug --loglevel verbose --debug-synchronization 1000",
```
markholland: 1000 means that anything that takes more than 1000ms to synchronize will be logged. Great way to see network latency being an issue.

zoeabryant:
Great idea to add screenshots!
when do the screenshots get cleared out? are we worried about storing too many?

markholland:
@zoeabryant Well, each CI build will have it's own artifacts. Even so, with how it's currently written, each run will overwrite the existing files so even when used locally there will only ever be as many screenshots as there are tests that take them. We can expand on this functionality in the future as we will want to ship screenshots to the store anyway.

e2e/init.js
```diff
@@ -2,7 +2,7 @@ import detox from "detox";
 import { detox as config } from "../package.json";
 
 // Set the default test timeout of 20s
```
RGBboy: This is now 200 seconds 😉

PR #227, opened 2018-05-04
Generate .env per job and use release keys for release jobs

markholland:
We want to use different API keys in our publicly distributed builds.

To accomplish this we override the required ENV_VARS from .env.example with our release keys.


markholland:
Production Contentful needs work before this can be merged

PR #232, opened 2018-05-04
Favouriting events: Unable to tap favourite on EventCard with VoiceOver enabled

### LINKED ISSUE #215: Favouriting events: With the Voiceover assistant enabled on iOS, it isn't possible to tap heart icon on the Events listing or Saved events page
## Description of issue

On iOS, with the Voiceover assistant enabled, it isn't possible to tap the heart icon to favourite/remove an event when viewing the Events listing page. It highlights the entire event card instead. 

## Expected behaviour

With Voiceover assistant enabled, tapping on the heart icon should select it and read out the label for its current state.  (Like it does for Android)

## Steps to reproduce

- Enable the Voiceover tool
- Go to the Events listing page
- Tap on the heart icon for one of the events
> The entire event card is selected

iPhone 7+

RGBboy:
Hopefully fixes #215.

It is not possible to have nested accessible elements. In this case we can not have the EventCard accessible and the nested SaveEventButton accessible too. 

**Note:** This PR disables accessibility on the Event Card, however the headings, time etc should still be accessible. This may not be the right approach so needs testing to determine best approach.

## Tester check-list

Accessibility Testing


declanslevin:
The event card itself is no longer accessible but now the various elements within it are individually accessible, including the favouriting icon. 

I think this has fixed the issue.

frigus02:
It feels a bit weird in TalkBack. Because the event card itself is no longer focusable, the screen reader never announces that this is a button.

We could try moving the save favorite button outside of the event card `Touchable`. Then position it above the card using `position: "absolute"`.

RGBboy:
@frigus02 Yeah, I wasn't completely sure how this would pan out. I have another idea that would make it better. Will try that soon.

RGBboy:
@declanslevin I have refactored to have a button with all the content on the left of the card, and a button with the save ability on the right. Hopefully it is a much better accessibility experience now. Would you mind retesting?

declanslevin:
Re-tested and I agree this is a more elegant solution which is working correctly. 

This can be merged once the conflicts are resolved.

Issue #100, opened 2018-04-06
Free/Price values not validated in app


declanslevin:
Tested with Galaxy S8 and iPhone 7+

# Description of issues

**Scenario 1**
If you reverse the price values for an event on Contentful e.g. 20 in Event Price Low and 10 in Event Price High, the app doesn't recognise this and reports "From: £20" on the event thumbnail and just "£20" on the details page. 

See "Mr. Poopybutthole" for an example event. 

**Scenario 2**
If an event is marked as free, but prices values are set in the Low/High fields, then the event displays "From: £20" on the event thumbnail, but "Free" on the event details page.

I can potential for this happening by accident when a content editor is creating an event in Contentful due to the fact that the price fields have a scrolling function which they may inadvertently modify without realising. 

See "The Imp" for an example event.

**Scenario 3**
If an event is marked as not free but 0 is put for both prices it will display £0.

See "Sasha Velour" for an example event. 

# Expected Behaviour

**Scenario 1** - If "Price Low" is greater than "Price High" then the price values are swapped.

**Scenario 2** - If the event is marked as "Free", but a price value is added in one/both of the price fields, it is marked as free. This may require some discussion though. Perhaps some validation could be added, so that if "Free" is set to yes, the price low/high fields _have_ to be 0 to publish changes. 

**Scenario 3** - If there is not a value greater than 0 in the price fields, the event is marked as free.

# Steps to reproduce

- Open one of the example events (Mr. Poopybutthole / The Imp / Sasha Velour)

# Screenshots

**Scenario 1**
<img width=400 src="https://user-images.githubusercontent.com/33623842/38991956-2b43b1d0-43d7-11e8-9917-99bcc34d2675.jpg">
<img width=400 src="https://user-images.githubusercontent.com/33623842/38991891-fcb29520-43d6-11e8-9c7c-1366ca7a7305.jpg">

**Scenario 2**
<img width=400 src="https://user-images.githubusercontent.com/33623842/39138581-2c8597d6-4718-11e8-9648-f40caf342d20.jpg">
<img width=400 src="https://user-images.githubusercontent.com/33623842/39138584-2f7f4428-4718-11e8-80e6-25d0690c3c5f.jpg">

**Scenario 3**
<img width=400 src="https://user-images.githubusercontent.com/33623842/38991973-383ed6bc-43d7-11e8-86a0-1daae20011c1.jpg">
[...]

PR #235, opened 2018-05-04
Free events should look free

zoeabryant:
For Scenarios 2 and 3 in https://github.com/redbadger/pride-london-app/issues/100  

- Ignore 'High' and 'Low' pricing if the event is marked as 'Free'
- If 'High' and 'Low' pricing are both '£0' the event should be treated as if it is marked 'Free'

## Pre-flight check-list

Before raising a pull request

* [ ] Documentation
* [x] Unit tests

## Pre-merge check-list

* [x] Link to Trello ticket/GitHub issue (If applicable) https://github.com/redbadger/pride-london-app/issues/100  
* [ ] Test plan provided
* [ ] Tester approved

## Tester check-list

  * [ ] Galaxy S7 (Android 6.0)
  * [ ] iPhone X (iOS 11.x)
  * [ ] iPhone 8 (11.x)
  * [ ] iPhone 7 (iOS 10.x)

  Optional:

  * [ ] Google Pixel 2
  * [ ] Galaxy S8 Plus
  * [ ] Galaxy S7 Edge
  * [ ] Galaxy S6
  * [ ] iPhone 7+
  * [ ] iPhone 6

* [ ] Large Text (=<200%)

* [ ] Airplane mode
* [ ] 2G/3G Network Settings


charypar:
I'd try to reduce the redundancy of price information:

1. When fetching data, look at the three values and set the prices we'll use in the app, free events having a price of 0 - 0. We can also swap the prices in case they are the wrong way around. This then means we don't have to drag around the `isFree` flag and reproduce the logic every time we're deciding whether an event is free or not.
2. In the price formatting function, which we hopefully use everywhere when displaying a price, we can just make the exception that if both prices are 0, we display "Free"

Does that make sense?

RGBboy:
@charypar I would suggest we drop the isFree input from contentful too then. We can add help text to the Event Price Low field to call this out.

zoeabryant:
Closed - now working in https://github.com/redbadger/pride-london-app/pull/257 with @matt-paul

PR #243, opened 2018-05-05
156 Header banner

frigus02:
[Trello Ticket](https://trello.com/c/NAgFj4ia/156-home-header-banner)

Android | iOS
--- | ---
image | image

## Test Plan

* Create a few header banner items in contentful.
* The start screen should show one of them.
* The banner should change every day.
* Landscape should work as well.

## Pre-flight check-list

Before raising a pull request

* [x] Documentation
* [x] Unit tests

## Pre-merge check-list

* ~Any risky areas identified~
* [x] Test plan provided
* [ ] Tester approved

## Tester check-list

  * [ ] Galaxy S7 (Android 6.0)
  * [ ] iPhone X (iOS 11.x)
  * [ ] iPhone 8 (11.x)
  * [ ] iPhone 7 (iOS 10.x)

  Optional:

  * [ ] Google Pixel 2
  * [ ] Galaxy S8 Plus
  * [ ] Galaxy S7 Edge
  * [ ] Galaxy S6
  * [ ] iPhone 7+
  * [ ] iPhone 6

* [ ] Large Text (=<200%)

* [ ] Airplane mode
* [ ] 2G/3G Network Settings


RGBboy:
Just a heads up, the header does not immediately show if you have an old version of the contentful data. Do we have a ticket or issue raised to fetch data at start up?

RGBboy:
Here are a couple of screenshots showing the header background color problem. This only happens on the home screen as other screens have a solid purple background behind the status bar. We just need to consider that on the home screen the contract changes from dark to light. Note: The header banner changes to a number of different colors too so this will need to be considered in the design.

screen shot 2018-05-08 at 10 16 08
screen shot 2018-05-08 at 10 16 22

RGBboy:
Also note: This content will need to be added to the prod contentful. When you are happy with this I don't mind adding it.

frigus02:
We can easily change the status bar text/icon color. Not sure how easy it is to do this automatically based on the background color. But at least another field on contentful would work.

We still have the issue of what happens when the user scrolls down. Do we keep the status bar in the background color from the image?

RGBboy:
From chatting with the design team I believe they are thinking of setting a solid color background on the home page to that of the image background color, on other pages it would be purple. I suggested pairing with a dev to resolve this though as it may not be the result they really want.

RGBboy:
Merged, even though the coverage looked bad. Essentially it is because this PR considered all our connector components (I.e. screen index.js files) that we do not currently test.

frigus02:
Yeah, I noticed that. I guess it's a good thing these appear in the coverage now. Makes us aware of the fact that the redux connectors are not covered.

PR #246, opened 2018-05-07
46 (Part 2) - Recurring events in event list

robbiemccorkell:
## Pre-flight check-list

Before raising a pull request

* [ ] Documentation
* [ ] Unit tests

## Pre-merge check-list

* [ ] Test plan provided
* [ ] Tester approved

## Tester check-list

  * [ ] Galaxy S7 (Android 6.0)
  * [ ] iPhone X (iOS 11.x)
  * [ ] iPhone 8 (11.x)
  * [ ] iPhone 7 (iOS 10.x)

  Optional:

  * [ ] Google Pixel 2
  * [ ] Galaxy S8 Plus
  * [ ] Galaxy S7 Edge
  * [ ] Galaxy S6
  * [ ] iPhone 7+
  * [ ] iPhone 6

* [ ] Large Text (=<200%)

* [ ] Airplane mode
* [ ] 2G/3G Network Settings


robbiemccorkell:
I think I'm done. This is now ready for review

robbiemccorkell:
I think I'm done. This is now ready for review

src/data/formatters.js
```diff
@@ -16,6 +16,19 @@ export const formatDateRange = (dateRange: DateRange) =>
...
 
+export const formatContentfulDate = (
+  year: string,
+  month: string,
+  day: string,
+  time?: string
+) => {
+  const correctedDay = day.length < 2 ? `0${day}` : day;
+  const correctedMonth = month.length < 2 ? `0${month}` : month;
+  const correctedYear = year.length < 4 ? `20${year}` : year;
```
frigus02: FYI, JavaScript has [`String.prototype.padStart`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/padStart) now, so this could be:
```js
day.padStart(2, "0")
month.padStart(2, "0")
year.padStart(4, "2000")
```

This is available in Node.js 9 and all browsers. Not sure what we can use in React Native, though.

src/selectors/events.js
```diff
@@ -50,6 +52,77 @@ export const groupEventsByStartTime = (events: Event[]): EventDays => {
...
+      },
+      endTime: {
+        [locale]: startAndEndAreSameDay
+          ? formatContentfulDate(
+              recurranceYear,
+              recurrancyMonth,
+              recurranceDay,
+              eventEndTime
+            )
+          : event.fields.endTime[locale]
```
frigus02: Can you explain the behaviour when an `startTime` and `endTime` of an event are not on the same day and it has recurrences? From the code it looks like the recurrence then ends on the `endTime` of the event. What would be an example of an event where this is useful?

Can it happen that the recurrence then starts after it ends? This crashed our app previously for normal events. Maybe we need a guard here.

robbiemccorkell: Yea this was a bit of an odd one, and definitely has some edge cases. I figured that if the end time was on same day then we would definitely want to change the day to that of the recurrence.

However I figured if the content editor has selected the event dates over a wide range of days (which they really shouldn't if they are using the recurring days feature) then it would be best not to mess with it.

I guess it is possible that you could end up with an end date that is before the start day of a recurrence with this logic. Maybe I'll tweak the logic so if the end date is the same day _or_ a previous day then modify the day.

src/data/formatters.js
robbiemccorkell: Oh cool. I'll try it out.

PR #258, opened 2018-05-09
Integrate with Bugsnag for crash reporting

markholland:
In order to capture crashes from the app we have chosen Bugsnag as it can report both native and js crashes.

- [x] Integrate and use test crash to validate
- [x] Upload mapping files as part of CI jobs [link](https://docs.bugsnag.com/platforms/react-native/showing-full-stacktraces/)
- [x] Distinguish between our alpha/beta/release channels [link](https://docs.bugsnag.com/platforms/react-native/configuration-options/#releasestage)
- [x] Integrate with Slack

Example crash report of error thrown in js with source maps uploaded [link](https://app.bugsnag.com/pride-in-london/pride-in-london-2018-app/errors/5af40cc35e88900017444a88?filters%5Bevent.since%5D%5B%5D=30d&filters%5Berror.status%5D%5B%5D=open&event_id=5af40cc35e88900017444a87)

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -9,9 +9,12 @@ import SplashScreen from "react-native-splash-screen";
 import Config from "react-native-config";
 import { Client, Configuration } from "bugsnag-react-native";
 
+import analytics from "./integrations/analytics";
 import reducers from "./reducers";
+import { init } from "./actions";
 import { getEvents } from "./actions/events";
 import { loadSavedEvents } from "./actions/saved-events";
+import { navigate } from "./actions/navigation";
 import App from "./App";
 
 const bugsnagConfiguration = new Configuration();
@@ -30,12 +33,15 @@ YellowBox.ignoreWarnings([
 
 const store = createStore(
   reducers,
-  composeWithDevTools(applyMiddleware(thunk))
+  composeWithDevTools(applyMiddleware(thunk, analytics))
 );
+
+const handleNavigationChange = navigate(store.dispatch);
 
 class AppWrapper extends Component<{}> {
   componentDidMount() {
     SplashScreen.hide();
+    store.dispatch(init());
     store.dispatch(getEvents());
     store.dispatch(loadSavedEvents());
   }
@@ -43,7 +49,7 @@ class AppWrapper extends Component<{}> {
   render() {
     return (
       <Provider store={store}>
-        <App />
+        <App onNavigationStateChange={handleNavigationChange} />
       </Provider>
     );
   }
--- a/android/app/src/main/java/org/prideinlondon/festival/Analytics.java
+++ b/android/app/src/main/java/org/prideinlondon/festival/Analytics.java
@@ -1,0 +1,51 @@
+package org.prideinlondon.festival;
+
+import com.crashlytics.android.answers.Answers;
+import com.crashlytics.android.answers.CustomEvent;
+
+import com.facebook.react.bridge.NativeModule;
+import com.facebook.react.bridge.ReactApplicationContext;
+import com.facebook.react.bridge.ReactContext;
+import com.facebook.react.bridge.ReactContextBaseJavaModule;
+import com.facebook.react.bridge.ReactMethod;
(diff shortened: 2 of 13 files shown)


.circleci/config.yml
```diff
@@ -197,6 +197,38 @@ release-api-keys: &release-api-keys
       echo 'export GOOGLE_MAPS_API_KEY=$RELEASE_GOOGLE_MAPS_API_KEY' >> $BASH_ENV
       source $BASH_ENV
 
+set-beta-stage: &set-beta-stage
```
markholland: We set a stage env_var for bugsnag to use, this way we'll be able to filter in the bugsnag dashboard by beta and release.

.circleci/config.yml
```diff
@@ -289,6 +324,7 @@ jobs:
                 fabric_build_secret:$FABRIC_BUILD_SECRET \
                 fabric_groups:$FABRIC_GROUPS \
                 commit_sha:$CIRCLE_SHA1
+      - <<: *bugsnag-upload-sourcemaps-android
```
markholland: Here we upload the js sourcemaps. They differ per platform because React Native

.circleci/config.yml
```diff
@@ -378,7 +420,10 @@ jobs:
                 fabric_build_secret:$FABRIC_BUILD_SECRET \
                 fabric_groups:$FABRIC_GROUPS \
                 ci_build_num:$CIRCLE_BUILD_NUM \
-                commit_sha:$CIRCLE_SHA1
+                commit_sha:$CIRCLE_SHA1 \
```
markholland: We also have to upload 'sourcemaps' for iOS, this is handled inside the Fastlane file

android/build.gradle
```diff
@@ -32,5 +32,8 @@ allprojects {
             // All of React Native (JS, Obj-C sources, Android binaries) is installed from npm
             url "$rootDir/../node_modules/react-native/android"
         }
+        maven {
```
markholland: Part of bugsnag setup docs

ios/fastlane/Fastfile
```diff
@@ -98,6 +103,12 @@ platform :ios do
     )
     if options[:submit]
       deploy_testflight
+      download_dsyms(version: get_version_number)
```
markholland: For release builds the debug symobls come from Apple because we only upload bitcode, the final compilation happens on their side.

package.json
```diff
@@ -52,6 +53,7 @@
     "babel-eslint": "^8.2.1",
     "babel-jest": "22.2.2",
     "babel-preset-react-native": "4.0.0",
+    "bugsnag-sourcemaps": "^1.0.3",
```
markholland: Added this as a dev depdendency so that it can be easily cached in CI

android/build.gradle
frigus02: We already have this by using `google()`, see https://developer.android.com/studio/build/dependencies#google-maven. Also as you can see there `google()` is relatively new. So I assume the Bugsnag docs have not been updated, yet.

src/index.js
```diff
@@ -6,11 +6,22 @@ import { Provider } from "react-redux";
...
 import { loadSavedEvents } from "./actions/saved-events";
 import App from "./App";
 
+const bugsnagConfiguration = new Configuration();
+bugsnagConfiguration.releaseStage = Config.RELEASE_STAGE
+  ? Config.RELEASE_STAGE
+  : "dev";
+// Only send reports for releases from master branch
+bugsnagConfiguration.notifyReleaseStages = ["beta", "release"];
+const bugsnag = new Client(bugsnagConfiguration); // eslint-disable-line no-unused-vars
```
frigus02: Do we need to do anything with this client after we created it?

markholland: Not for unhandled errors https://docs.bugsnag.com/platforms/react-native/

Eventually we should probably have some handled errors around contentful fetching

Issue #241, opened 2018-05-05
Tapping an event multiple times causes navigation to stack it more than one


charypar:
## Description of issue

When on home screen, tap one of the featured events multiple times in quick succession. The app navigates to the event as many times as you tapped, and it take as many taps on "back" to get back.

## Expected behaviour

App navigates only once, other taps are ignored.

## Steps to reproduce

1. Open home screene
2. _quickly_ tap on an event multiple times

## OS / Browser / Device model

iPhone 7, iOS 11.1 - probably affects all devices and versions

PR #266, opened 2018-05-09
241 multiple nav stack

abinocoe:
Fixing issue of duplicate screens being added to the navigation stack on rapid event taps
issue here: https://github.com/redbadger/pride-london-app/issues/241

Was solved by upgrading react-navigation package to v2. 
This required minimal code changes, only deprecation of previous navigation creator functions which have been replaced by new mostly identical functions - save for the tabNavigator creator.

Rendering of the TabBar was broken upon upgrade, fixing this uncovered that the Animated components for the icons and labels weren't being used so have swapped them out for their basic counterparts.

  * [ ] Galaxy S7 (Android 6.0)
  * [ ] iPhone X (iOS 11.x)
  * [ ] iPhone 8 (11.x)
  * [ ] iPhone 7 (iOS 10.x)

  Optional:

  * [ ] Google Pixel 2
  * [ ] Galaxy S8 Plus
  * [ ] Galaxy S7 Edge
  * [ ] Galaxy S6
  * [ ] iPhone 7+
  * [ ] iPhone 6


src/Navigation.js
```diff
@@ -81,8 +84,6 @@ const HomeStack = StackNavigator(
   {
     initialRouteName: HOME,
     navigationOptions: {
-      tabBarIcon: tabIcon(iconHomeDefault, iconHomeActive),
-      tabBarLabel: text.tabHome,
       tabBarTestIDProps: {
```
RGBboy: This needs to be moved down into the tabBarNavigator props (like the others). I think this will help the e2e tests pass.

src/Navigation.js
```diff
@@ -91,77 +92,92 @@ const HomeStack = StackNavigator(
...
-    },
     headerMode: "none"
   }
 );
 
-const ParadeStack = StackNavigator(
+const ParadeStack = createStackNavigator(
   {
-    [PARADE]: { screen: withShadow(ParadeInformationScreen) }
+    [PARADE]: { screen: withShadow(() => <View />) }
```
RGBboy: This line needs to stay the same for the ParadeInformationScreen to show. (It was added while you were developing this)

src/Navigation.js
abinocoe: Done!

src/Navigation.js
abinocoe: Also done!

src/components/NavigationTabBar.js
```diff
@@ -61,67 +61,44 @@ class NavigationTabBar extends React.PureComponent<_TabBarBottomProps> {
...
-      inputRange,
-      outputRange: inputRange.map(i => (i === index ? 1 : 0))
-    });
-    const inactiveOpacity = position.interpolate({
-      inputRange,
-      outputRange: inputRange.map(i => (i === index ? 0 : 1))
-    });
+    const { route, index, focused } = scene;
+    const opacity = focused ? 1 : 0;
+    const opacitInv = focused ? 0 : 1;
```
frigus02: Oh nooooo, the animated `position` doesn't exist anymore? I think I was able to see/feel this animation of the opacity 😊.

RGBboy: Yeah it's gone. We can roll our own though. We just could not see the difference so took it out.

RGBboy:
With respect to the `headerMode:"none"` problem, I believe static properties on screen components no longer works. I think there is a way to declare this within the navigation setup though.

RGBboy:
I would suggest we get test to double check this too before it goes out, just in case we missed something.

frigus02:
Okay, `headerMode` is obviously working just fine 🙈. What I meant is:

```js
static navigationOptions = {
  tabBarVisible: false
};
```

abinocoe:
the `tabBarVisible` issue has now been addressed, but I'd love it if anyone has any advice if there's a cleaner way to do it

src/components/NavigationTabBar.js
abinocoe: I've sorted the `tabBarVisible` issue, would love to hear if you have any ideas as to cleaner alternatives to my solution :)

src/Navigation.js
```diff
@@ -147,38 +162,84 @@ const SavedStack = StackNavigator(
...
+      screen: EventsStack,
+      navigationOptions: ({ navigation }) =>
+        getNavigationOptions(
+          navigation,
+          {
+            tabBarIcon: tabIcon(iconEventsDefault, iconEventsActive),
+            tabBarLabel: text.tabEvents,
+            tabBarVisible: true
+          },
+          [EVENT_DETAILS]
```
frigus02: The route `EVENT_CATEGORIES_FILTER` shouldn't have a taskbar, either 🙂

frigus02:
I think we only want the tab bar to be visible on the initial route of a tab. On all sub routes it should be hidden. So maybe something like this would make it a bit simpler:

```js
import type {
  NavigationScreenConfigProps,
  NavigationTabScreenOptions
} from "react-navigation";

const hideTabBarOnSubRoutes = (
  initialRouteName: string,
  navigationOptions: NavigationTabScreenOptions
) => ({ navigation }: NavigationScreenConfigProps) => {
  const { routeName } = navigation.state.routes[navigation.state.index];
  return {
    ...navigationOptions,
    tabBarVisible: routeName === initialRouteName
  };
};

const TabNav = createBottomTabNavigator(
  {
    [HOME]: {
      screen: HomeStack,
      navigationOptions: hideTabBarOnSubRoutes(HOME, {
        tabBarIcon: tabIcon(iconHomeDefault, iconHomeActive),
        tabBarLabel: text.tabHome
      })
    }
  }
);
```

It would make flow happy, too, as far as I can see.

abinocoe:
I love this, thank you! I went through all the screens to check for the
static nav option of tabBarVisible set to false but I guess my
in-familiarity with the app tripped me up in realising that it was only
meant for the initial routes :)

frigus02:
No worries. I believe we changed this mid-way, so now the tab bar is hidden for all sub routes. But you might want to clearify this with someone from the design/UX team 🙂.

PR #278, opened 2018-05-11
e2e coverage for tab navigation

markholland:
e2e is a lot more stable than before so starting to add coverage for some low hanging fruit


RGBboy:
Just FYI, #266 is coming in hot! react-navigation v2. There are a few changes that will mean the test ids need to be declared on the tab navigator rather than the individual stack navigators.

Issue #223, opened 2018-05-03
Favouriting Events: Saved event icons flicker after rotating device to landscape and back to portrait (iOS only)


declanslevin:
## Description of issue

When you are on the Events listing page, if you are viewing several events that have been "saved", then their heart icons will flicker on and off a few times (before returning to the correct state) when rotating the device to landscape and back to portrait.  

## Expected behaviour

The heart icons don't flicker when rotating the device. 

## Steps to reproduce

- Open the Events listing page
- Tap to favourite the top 3 events
- Rotate device to landscape
- Rotate it back to portrait
> Heart icons flicker off and on a few times before returning to correct state. 

## OS / Browser / Device model

iPhone 6 (10.3.3)
iPhone 7+ (10.3.1)

## Screenshots

Couldn't get a screen recording, but can provide a demonstration if needed.

RGBboy:
Closing as this should be fixed by @282

Issue #221, opened 2018-05-03
Favouriting events: Saving/removing an event or scrolling event listings page causes heart icons to flicker (iOS only)


declanslevin:
## Description of issue

When on the Events listing page, tapping to add or remove an event from the saved events will cause the heart icons to flicker 

## Expected behaviour

Tapping to add/remove an event doesn't affect the icons for other saved events. 

## Steps to reproduce

- Open the "Events" page
- Favourite the top 2 events
- Repeatedly tap the heart icon for the third event
> The heart icons for the top 2 events flicker as the icon for the third event is tapped. 

## OS / Browser / Device model

iPhone 6 (10.3.3)
iPhone 7+ (10.3.1)

## Screenshots

![fave_flicker](https://user-images.githubusercontent.com/33623842/39579160-fa5f5792-4edd-11e8-9590-7a80792c2ca6.gif)

-----

I also noticed that scrolling with a bunch of favourited events visible in the events listing page will also result in the icons flickering:

![scroll_flicker](https://user-images.githubusercontent.com/33623842/39583723-abfa9656-4ee8-11e8-826b-2ac81902832d.gif)

Again, it only affects iOS so I assume it is related to the above issue - if not I can log it as a separate issue. 

It is quite easy to reproduce on iPhone 6 - I have reproduced on iPhone 7+ once or twice but it's not reliably reproducible.

RGBboy:
@declanslevin I could not reproduce this on master. There could be a chance someone has made a change to fix. Would you mind showing me this when you are in?

stevecottle:
Adding Accessibility label as the flash icons may have accessibility concerns.

RGBboy:
Fixed in #282

PR #282, opened 2018-05-11
Fix heart flicker on iOS

RGBboy:
Hopefully this change sorts out both #221 and #223

It just adds `shouldComponentUpdate` to the saveEventButton.

## Tester check-list

* [ ] iPhone 6 (iOS 10.x)


declanslevin:
This looks good on iPhone 6 and 7+

PR #288, opened 2018-05-12
Fix text resizing

frigus02:
## Pre-flight check-list

Before raising a pull request

* [ ] Documentation
* [ ] Unit tests

## Pre-merge check-list

* [ ] Test plan provided
* [ ] Tester approved

## Tester check-list

  * [ ] Galaxy S7 (Android 6.0)
  * [ ] iPhone X (iOS 11.x)
  * [ ] iPhone 8 (11.x)
  * [ ] iPhone 7 (iOS 10.x)

  Optional:

  * [ ] Google Pixel 2
  * [ ] Galaxy S8 Plus
  * [ ] Galaxy S7 Edge
  * [ ] Galaxy S6
  * [ ] iPhone 7+
  * [ ] iPhone 6

* [ ] Large Text (=<200%)

* [ ] Airplane mode
* [ ] 2G/3G Network Settings

### DIFF
--- a/src/components/Text.js
+++ b/src/components/Text.js
@@ -63,12 +63,15 @@ class Text extends React.PureComponent<Props> {
       />
     ) : (
       <RnText
+        allowFontScaling={false}
         style={[styles[typedType], styles[typedColor], style]}
         {...otherProps}
       />
     );
   }
 }
+
+const cap = (def, max) => Math.min(max, def * PixelRatio.getFontScale());
 
 const textStyles = {
   uber: {
@@ -85,41 +88,41 @@ const textStyles = {
   },
   h2: {
     fontFamily: "Poppins-SemiBold",
-    fontSize: 18,
-    lineHeight: 24,
+    fontSize: cap(18, 22),
+    lineHeight: cap(24, 29),
     includeFontPadding: false
   },
   h3: {
     fontFamily: "Poppins-SemiBold",
-    fontSize: 16,
-    lineHeight: 20,
+    fontSize: cap(16, 20),
+    lineHeight: cap(20, 25),
     includeFontPadding: false
   },
   h4: {
     fontFamily: "Roboto-Medium",
-    fontSize: 16,
-    lineHeight: 24
+    fontSize: cap(16, 20),
+    lineHeight: cap(24, 30)
   },
   text: {
     fontFamily: "Roboto",
-    fontSize: 16,
-    lineHeight: 24
+    fontSize: cap(16, 20),
+    lineHeight: cap(24, 30)
   },
   small: {
     fontFamily: "Roboto",
-    fontSize: 14,
-    lineHeight: 20
+    fontSize: cap(14, 18),
+    lineHeight: cap(20, 26)
   },
   xSmall: {
     fontFamily: "Poppins-SemiBold",
-    fontSize: 12,
-    lineHeight: 16,
+    fontSize: cap(12, 14),
+    lineHeight: cap(16, 18),
     includeFontPadding: false
   },
   price: {
     fontFamily: "Roboto-Bold",
-    fontSize: 14,
-    lineHeight: 20
+    fontSize: cap(14, 18),
+    lineHeight: cap(20, 26)
   },
   blackColor: {
     color: blackColor
--- a/src/components/ButtonWithShapes.js
+++ b/src/components/ButtonWithShapes.js
@@ -109,7 +109,7 @@ class SupportUsButton extends React.PureComponent<Props> {
 const styles = StyleSheet.create({
   button: {
     justifyContent: "center",
-    height: 100 * PixelRatio.getFontScale(),
+    height: 100 * Math.min(1.25, PixelRatio.getFontScale()),
     minHeight: 100,
     borderRadius: 5,
     overflow: "hidden",
@@ -148,7 +148,7 @@ const styles = StyleSheet.create({
   titleIcon: {
     marginLeft: 8,
     alignSelf: "center",
-    transform: [{ translateY: -4 * PixelRatio.getFontScale() }]
+    transform: [{ translateY: -4 * Math.min(1.25, PixelRatio.getFontScale()) }]
   },
   text: {
     color: whiteColor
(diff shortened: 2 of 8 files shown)


RGBboy:
With respect to Markdown, we have tried to limit its use to just paragraphs and inline styles for now (e.g. bold, emphasis). It shouldn't end up being used for headings or other kinds of typographic content as these have not been designed for.

frigus02:
Yes, but as soon as we have the markdown renderer library, we loose control over font scaling in these pieces of text. Unfortunately.

RGBboy:
I looked into other markdown renderers too. There are not that many out there. I wonder if we could roll our own? Looking at the current renderers source code, I don't think it would be a massive task. Alternatively we could submit a PR to add font scaling to the lib we do use.

src/components/Text.js
```diff
@@ -154,7 +163,7 @@ const markdownDefaultStyles = {
     height: 4,
     backgroundColor: "black",
     borderRadius: 2,
-    marginTop: 11 * PixelRatio.getFontScale(),
+    marginTop: 10 * Math.min(1.29, PixelRatio.getFontScale()),
```
RGBboy: is the `1.29` intentional? Or is this supposed to be `1.25`?

src/screens/SponsorScreen/__snapshots__/component.test.js.snap
```diff
@@ -173,8 +173,8 @@ exports[`SponsorScreen Component renders correctly 1`] = `
           Object {
             "text": Object {
               "color": "#2d2f7f",
-              "fontSize": 14,
-              "lineHeight": 20,
+              "fontSize": 9,
+              "lineHeight": 13,
```
RGBboy: Intentional? I think the change is right, but not 100%.

Also, I was thinking about this. Should we just make a component to do this rather than going through markdown? This is a custom look (small purple text with bullets) and I think we might be better off just creating an UnorderedList and ListItem component instead.

src/screens/DonateScreen/SegmentedControl.js
```diff
@@ -69,7 +69,7 @@ const styles = StyleSheet.create({
     borderColor: mediumGreyColor,
     borderTopWidth: 1,
     borderBottomWidth: 1,
-    height: 50 * PixelRatio.getFontScale(),
+    height: 50 * Math.min(1.22, PixelRatio.getFontScale()),
```
RGBboy: `1.22`? Keen to understand where this numbers came from. If this is intentional, I wonder if there is a more meaningful way for us to calculate this. I.e. some relation to `lineHeight`? or `fontSize`?

src/components/__snapshots__/Text.test.js.snap
```diff
@@ -10,26 +10,26 @@ exports[`does not render markdown images 1`] = `
       },
       "h1": Object {
         "fontFamily": "Poppins-Bold",
-        "fontSize": 24,
+        "fontSize": 12,
```
RGBboy: I'm a little concerned that by default these get returned as half of the intended font size in our tests. Is there a reason this is happening?

src/components/FilterHeaderCategories.js
```diff
@@ -92,7 +92,7 @@ const styles = StyleSheet.create({
     backgroundColor: categoriesFilterButtonBgColor,
     alignItems: "center",
     width: 44,
-    height: 44 * PixelRatio.getFontScale(),
+    height: 44 * Math.min(1.25, PixelRatio.getFontScale()),
```
RGBboy: With `44 * Math.min(1.25, PixelRatio.getFontScale())`, would you mind creating a helper with a meaningful name? It is not immediately obvious what the intention is here and a named function could help that.

frigus02: Yes, that's a good idea.

src/screens/DonateScreen/SegmentedControl.js
frigus02: Yes, that's intentional. But we definitely need a way to make this understandable.

In this case this style is for the button, which is an `h2`. The `h2` has a default font size of 18 and a max size of 22.

We want the height to cap at the same scale. Result: 22 / 18 = 1.22

frigus02:
Looking at this again, I feel that this is really hard to understand. The markdown library is an issue, but I don't think it's the biggest one.

We have a lot of places where we define the height of a component based on the font scale. These now need to know about the text style used for their children (h1, h2, text, small, ...) because all of them have different scale caps.

I can see different solutions here:
*   Provide a function like `scaleWithText`, which can be used for styles:
    ```js
    button: {
      height: scaleWithText({ base: 44, textType: 'h2' })
    }
    ```
*   Make the scaling cap consistent for all text types. It seems that `1.25` should work pretty well. Then this could be a constant and the styles would look like this:
    ```js
    button: {
      height: 44 * Math.min(FONT_SIZE_CAP, PixelRatio.getFontScale())
    }
    ```

@RGBboy, what do you think?

src/screens/SponsorScreen/__snapshots__/component.test.js.snap
frigus02: Intentional, yes. And I think you are right about the list. A custom list component would be simpler.

RGBboy:
The first option looks great. I like the fact that we capture the intent in the name of the function to clarify why we are scaling the height. I would just add that as a helper exported from our Text component. It will also give us a central place to change this as we come to understand the little tweaks that may need to be made.

frigus02:
Since this is an explicit mock of `PixelRatio` in react native I don't mind this so much. It does show that font scaling works for us. But it is also quite confusing. I can have a look at making a custom mock for `PixelRatio`.

declanslevin:
Found a minor issue, #304. 

On iOS, if you change the font settings while the app is open, then you still get the huge text like previously. I'm not sure if this is really an issue though as I wouldn't expect users to do this very often...

Otherwise I think this looks good on the devices I've tested on:
Galaxy S8
iPhone 6
iPhone 7+

JunBadger:
Hey @frigus02 ! Are we still following these rules?

frigus02:
@JunBadger For the most part, yes. The only difference is xSmall. We have these values in the code currently:

Size | Default | Max
--- | --- | ---
xSmall | 12 | 14
price | 14 | 18
tarBarItem | 12 | 13

PR #292, opened 2018-05-13
Attempt to provide better accessibility for TextInput

### LINKED ISSUE #259: Donate page: "Other amount" field is awkward to use with Voice Assistant on Android
## Description of issue

When using the Voice Assistant on Android, it gets a bit awkward trying to interact with the "Other amount" field. 

Firstly it's a bit confusing when you tap on it: “Edit box, double tap to edit, double tap to enter text, double tap and hold to long press”

Secondly, to actually interact with it, you need to double tap and hold. This is quite cumbersome. iOS only requires a double tap. 

## Expected behaviour

When you single tap on the "Other amount" field, it reads out "Other amount edit box, double tap to edit".
Then when you double tap, you can edit the field. 

## Steps to reproduce

- Enable the Voice Assistant
- Open the Donation page
- Tap on the Other amount field
> Reads out "Edit box, double tap to edit, double tap to enter text, double tap and hold to long press"
- Double-tap on Other amount field
> Doesn't work to edit field

Galaxy S8

frigus02:
Fixes #259 

There are at least 2 issues with input fields with React Native on Android: you need to double tap and hold to focus them and there is no way to specify a label (see https://github.com/facebook/react-native/issues/14989).

This attempts to these things with a hack. That's why it needs proper testing on both Android and iOS.
* "Other amount" field on Donate page is now activatable with a double tap.
* It has a custom label, which mimics the default TalkBack behaviour for text fields.

I also added another fix, which I might split out into a separate PR. The radio buttons for predefined amounts now announce, when they have been checked.

## Pre-flight check-list

Before raising a pull request

* [ ] Documentation
* [ ] Unit tests

## Pre-merge check-list

* [ ] Test plan provided
* [ ] Tester approved

## Tester check-list

  * [ ] Galaxy S7 (Android 6.0)
  * [ ] iPhone X (iOS 11.x)
  * [ ] iPhone 8 (11.x)
  * [ ] iPhone 7 (iOS 10.x)

  Optional:

  * [ ] Google Pixel 2
  * [ ] Galaxy S8 Plus
  * [ ] Galaxy S7 Edge
  * [ ] Galaxy S6
  * [ ] iPhone 7+
  * [ ] iPhone 6

* [ ] Large Text (=<200%)

* [ ] Airplane mode
* [ ] 2G/3G Network Settings


src/components/Touchable.android.js
```diff
@@ -1,29 +1,49 @@
...
+  componentDidUpdate(prevProps: TouchableProps) {
+    // Notify user, that the radio button changed to checked state.
+    if (
+      this.props.accessibilityComponentType === "radiobutton_checked" &&
+      this.props.accessibilityComponentType !==
+        prevProps.accessibilityComponentType
+    ) {
+      UIManager.sendAccessibilityEvent(
+        findNodeHandle(this),
+        UIManager.AccessibilityEventTypes.typeViewClicked
```
frigus02: The code for this was taken straight from the [docs](http://facebook.github.io/react-native/docs/accessibility.html#sending-accessibility-events-android).

src/screens/DonateScreen/NumberTextField.js
```diff
@@ -35,25 +72,52 @@ class DonateScreen extends React.PureComponent<Props, State> {
...
+        </Text>
+        <TextInput
+          keyboardType="numeric"
+          onBlur={this.onBlur}
+          onFocus={this.onFocus}
+          placeholder={focused ? placeholder : null}
+          placeholderTextColor={mediumGreyColor}
+          style={[styles.input, focused && styles.inputFocused]}
+          underlineColorAndroid={transparent}
+          importantForAccessibility="no"
```
frigus02: For some reason the TextInput works much better on TalkBack, when you hide it from screen readers completely and put in inside a view, which is accessible.

src/screens/DonateScreen/NumberTextField.js
```diff
@@ -35,25 +72,52 @@ class DonateScreen extends React.PureComponent<Props, State> {
...
+          label,
+          otherProps.value,
+          focused
+        )}
+        style={style}
+      >
+        <Text
+          type="h4"
+          color="lightNavyBlueColor"
+          importantForAccessibility="no"
```
frigus02: The label is also hidden because it is included in the `accessibilityLabel` of the parent view.

declanslevin:
Looks good on both Android and iOS, however there is one minor thing.
The read-out of the text field when an amount has been entered is a little confusing, e.g.:
> Editing, 25, Edit box, Other amount pound, double-tap and hold to long press

Maybe the value (25) should come after "Edit box" or "Other amount £"?

declanslevin:
Another thing I noticed is when tapping on a selected radio button, it will read out "Radio button 20 pounds selected. Double-tap to toggle".

This is misleading as the radio buttons can't be toggled off by double-tapping, only by selecting a different radio button.

frigus02:
@declanslevin Thanks for testing this.

> Maybe the value (25) should come after "Edit box" or "Other amount £"?

This is what TalkBack says for a native input field in the default Contacts app. I think we should check what VoiceOver says for a default input field and then copy this for iOS here. This way users will hear what they are used to. Would you mind recording this on iOS?

> Another thing I noticed is when tapping on a selected radio button, it will read out "Radio button 20 pounds selected. Double-tap to toggle".

I agree, this is confusing. However this is what Android says even in native apps for native radio buttons. So at least it is consistent behavior. There is just one difference to native apps now: When activating (using double tap) an already checked radio button, TalkBack says "checked" again. I will try implement the same behavior for us as well. I would not change the general text here, though.

Issue #120, opened 2018-04-13
Category filter button doesn't have a label so it is read out as "button" by text-to-speech


declanslevin:
## Description of issue

When tapping the Category filter button with a text-to-speech tool it reads out "button" without a label to describe what the button does. 

## Expected behaviour

When tapped, the text-to-speech tool should read out "Button - Category filter" (Android) or "Category filter - Button" (iOS).

## Steps to reproduce

- Enable your device's text-to-speech function
- Go to the events page in the Pride App 
- Tap on the Category filter button
> It reads out "button" 

## OS / Browser / Device model

Galaxy S8 (Android 8.0)
iPhone 7+

zoeabryant:
<img width="411" alt="screen shot 2018-05-11 at 14 25 15" src="https://user-images.githubusercontent.com/5781504/39926459-3736b710-5527-11e8-8188-6cf306bc15e5.png">

<img width="426" alt="screen shot 2018-05-11 at 14 25 27" src="https://user-images.githubusercontent.com/5781504/39926460-374d679e-5527-11e8-97de-3c89f2cb565d.png">

<img width="55" alt="screen shot 2018-05-11 at 14 25 41" src="https://user-images.githubusercontent.com/5781504/39926461-3762b1b2-5527-11e8-9654-06a151a4104e.png">

zoeabryant:
I can't figure out why when you tap `Showing` it reads `Events` instead - we haven't set it as a label anywhere! 🤔

zoeabryant:
I have looked into adding the `Filtering by x, y, z...` label - it doesn't actually read that out instead of whichever category you picked. On the web I'd usually add some hidden text to be read out instead, but it doesn't look like that concept works on apps - you need to tap to read.

frigus02:
The word "Showing" is read as "Events" because of the `accessibilityLabel`:
https://github.com/redbadger/pride-london-app/blob/5728b7edabfef670ce59d00cdee3b07410ce41e5/src/components/FilterHeaderCategories.js#L52
I think this is wrong. I am not sure what is should read, though. Should it say "Showing" or maybe "Showing all event types"?

[...]

PR #314, opened 2018-05-17
Filters accessibility labels

abinocoe:
Overriding/adding accessibility labels to filter options in modal and category filter list
As accessibility options in react native for android particularly are slim in terms of element types, the decision was made to override the defaults and extend the accessibility label to include a custom type of 'checkbox'. 
Checked status of box is incorporated.
Word order returned is platform specific to match typical behaviour.

## Pre-merge check-list
extension of #120 
* [ ] Test plan provided
* [ ] Tester approved


src/components/CheckBox.js
```diff
@@ -15,12 +15,19 @@ type Props = {
   style?: ViewStyleProp
 };
 
+const getAccessibilityLabel = (label: string, checked: boolean) =>
```
markholland: Maybe we need and accessibility utils that would export these types of functions

RGBboy: Agreed, accessibility module that exported `checkboxAccessibilityLabel(label: string, checked: boolean)` would be great. Then we can use it for the ListItem too.

abinocoe: excuse my noob-ish-ness, but where would you put this?

declanslevin:
Tested on Galaxy S8 and iPhone 7+, looks good!

declanslevin:
Tested on Galaxy S8 and iPhone 7+. Looks good other than a weird iOS issue which has been logged as #322

src/components/CheckBox.js
abinocoe: Have moved it out to the data folder. LMK if it's better somewhere else.

src/data/accessibility.js
```diff
@@ -0,0 +1,12 @@
+// @flow
+import { Platform } from "react-native";
+
+const checkboxAccessibilityLabel = (label: string, checked: boolean) =>
+  Platform.select({
```
frigus02: Do we want to use 2 separate files (accessibility.ios.js and accessibility.android.js) instead of `Platform.select`? I think this is cleaner because then iOS does not need to load the Android code and vice versa.

abinocoe: If you think that is best? I'm happy to

PR #316, opened 2018-05-17
Fix filter by date

### LINKED ISSUE #308: Unable to filter by date in test environment
## Description of issue

In beta for version 1.0.7 (iOS 5424, android 5426)

(Or any build after 59c7986a6614c0e7ef93a35e946cd8ce93b5198d)

## Expected behaviour

Can use the calendar to select a date to filter by

## Steps to reproduce

Go to events screen.
Select filter by date.
Select any date.
Watch it 🔥 

All.

## Screenshots

screen shot 2018-05-16 at 09 40 53

RGBboy:
Fixes #308: Unable to filter by date in test environment

This fixes an issue where events with varying timezones could result in a recurrence having a startTime after its endTime. Subsequently a call to `areRangesOverlapping` would throw an exception when trying to render the calendar.

**Note:** This does change the logic of `expandRecurringEventsInEntries` to always transform the endTime of a recurrence to be the same relative distance from its start time as the original start and endTime, hence the test changes.

## Tester check-list

  * [x] Galaxy S7 (Android 6.0)
  * [x] iPhone X (iOS 11.x)
  * [ ] iPhone 8 (11.x)
  * [x] iPhone 7 (iOS 10.x)


robbiemccorkell:
We might need to just make it doubly clear in our content editor guide that if they want a single event to repeat on multiple days, use the recurring events field, and leave the start/end times to be just the event on a single day.

I was worried that someone might create recurring events, and set the endTime to represent the last day of the recurrences or something like that. But as we've found it opens a whole can of worms and different possibilities. So this does make things a lot simpler.

src/selectors/events.js
```diff
@@ -58,55 +55,50 @@ export const groupEventsByStartTime = (events: Event[]): EventDays => {
...
-    "T"
-  );
-  const eventEndTime = event.fields.endTime[locale].split("T")[1];
-  const [eventStartYear, eventStartMonth, eventStartDay] = eventStartDate.split(
-    "-"
+
+  const startTime = DateTime.fromFormat(
+    event.fields.startTime[locale],
+    contentfulFormat,
+    { setZone: true }
```
robbiemccorkell: How do these options compare to using `DateTime.fromISO(<contentfulString>)`? On my `luxon` branch this seems to work, but I probably missed some detail

robbiemccorkell: Hmm ok. I'll bear this all in mind for my `luxon` branch. Also I found we can make our tests more predictable by setting the env var `TZ=UTC` on a test run. Which makes changing the tests a more viable option. I'll do that on my branch too.

robbiemccorkell:
@RGBboy A good example of start/end being across multiple days, as well as having recurring days set, is the "Lonely Planet" event in the production contentful.

RGBboy:
@robbiemccorkell Yeah I just saw that. We have updated the content editors guide to call out the expectation that start and end time should be for a single occurance.

robbiemccorkell:
👌

declanslevin:
Looks good

PR #317, opened 2018-05-18
Fix analytics screen event tracking

### LINKED ISSUE #275: Events analytics: Only the Home screen event is being sent from the beta
## Description of issue

Only the HOME screen event is being sent correctly from beta builds for Android and iOS. 

## Expected behaviour

Events for every screen are sent when the page is loaded, i.e. EVENT_LIST, EVENT_DETAILS, PARADE, etc. 

## Steps to reproduce

- Open the app
- View a bunch of different pages, i.e. Events, Parade, Saved, Support Us, etc. 
- Quit the app and restart it
- View "Screen Event" under the Events tab in Fabric
> It only shows data for the HOME page

Galaxy S8
iPhone 7+

RGBboy:
Fixes #275

TLDR; Analytics stopped working because App got refactored and missed passing the `onNavigationStateChange` to the Navigation stack. This PR fixes that.

## Tester check-list

  * [ ] Galaxy S7 (Android 6.0)
  * [ ] iPhone X (iOS 11.x)
  * [ ] iPhone 8 (11.x)
  * [ ] iPhone 7 (iOS 10.x)


RGBboy:
Jut double checked the iOS alpha build, and screen events are now coming through.

PR #318, opened 2018-05-18
Distribute all master builds externally

markholland:
Currently when we merge to master a build is pushed to iTunes Connect.

Unless we manually go in and assign new builds to external test groups then they will never receive any updates.

This changes how we submit builds to iTunes Connect to instead submit all builds to external distribution review but without notifying users of updates. Not notifying is to avoid spamming.


robbiemccorkell:
Awesome! I hadn't realised we could do this

markholland:
Still conflicted on when to merge this. It would mean external iOS testers get spammed every time we merge anything to master. I like the work @frigus02 has been doing towards scheduling jobs but I'm still unsure what the sweet spot would be, daily still seems too often. It would be great to reach a point where an external Beta is pushed and then x days later it's automatically promoted to production.

markholland:
I don't think is usable, once you submit a build to TestFlight review you can't submit any new builds until the first one has been reviewed. This means we'd get blocked almost instantly as review is at least a day.

PR #319, opened 2018-05-18
Google Maps in app on iOS

markholland:
We will need to use Google Maps in the app if we want to overlay information.

We used CocoaPods to assist with the initial setup and have decided to go with a recommended approach of checking in the Pods folder. This means we won't have to introduce Cocoapods to our dev or CI workflows which can cause headaches around versioning and greatly increase CI build time.

## Pre-merge check-list

* [ ] Test plan provided
* [ ] Tester approved

## Tester check-list

  * [ ] Galaxy S7 (Android 6.0)
  * [ ] iPhone X (iOS 11.x)
  * [ ] iPhone 8 (11.x)
  * [ ] iPhone 7 (iOS 10.x)

  Optional:

  * [ ] Google Pixel 2
  * [ ] Galaxy S8 Plus
  * [ ] Galaxy S7 Edge
  * [ ] Galaxy S6
  * [ ] iPhone 7+
  * [ ] iPhone 6

* [ ] Large Text (=<200%)

* [ ] Airplane mode
* [ ] 2G/3G Network Settings


ios/Podfile
```diff
@@ -0,0 +1,6 @@
+platform :ios, '9.0'
```
markholland: This is our Podfile which we can add to

src/screens/EventDetailsScreen/EventMap.js
```diff
@@ -19,6 +19,7 @@ const EventMap = ({ lat, lon, locationName }: Props) => (
   >
     <Delayed delay={600}>
       <MapView
+        provider={PROVIDER_GOOGLE}
```
markholland: We use Google Maps always

ios/PrideLondonApp/AppDelegate.m
```diff
@@ -28,6 +31,8 @@ - (BOOL)application:(UIApplication *)application didFinishLaunchingWithOptions:(
   
   [BugsnagReactNative start];
   
+  [GMSServices provideAPIKey:[ReactNativeConfig envFor:@"GOOGLE_MAPS_API_KEY"]];
```
markholland: We set the API key for Google Maps from our .env

robbiemccorkell:
I was never sure what to do with the cocoapods files. Is it normal to just commit them all, or do people usually have an install step in their build pipeline?

markholland:
@robbiemccorkell http://guides.cocoapods.org/using/using-cocoapods.html#benefits-of-checking-in-the-pods-directory

declanslevin:
The maps preview on the Event details page only loads correctly on the first Event opened, thereafter it appears as a pale blue colour:

markholland:
Great https://github.com/react-community/react-native-maps/issues/1452

markholland:
@declanslevin should be fixed in more recent builds

markholland:
I'll want to try and do this without CocoaPods as it has significantly increased build times. But this can be a back up should we need to start developing against Google Maps.

robbiemccorkell:
Out of interest, how do we know that overlays aren't supported in Apple Maps by react native maps? I hadn't seen that limitation.

markholland:
My understanding was that the overlay would be created in a google maps account and then the app would just refer to it.

markholland:
No need to merge this until we need it of course

declanslevin:
Retested on multiple iOS devices, map previews are appearing correctly. Looks good, I won't merge this for now though due to the previous comments.

markholland:
Have spoken with @frigus02 and this hopefully won't be needed. We should definitely try and use just the capabilities of react-native-maps. Also we should look into a feature of google developer console where you don't have to embed the google maps api key in the app. Would mean any key revokes wouldn't require app updates.

Will close this PR but keep the branch

PR #320, opened 2018-05-18
refresh events on foregrounding

abinocoe:
Addresses this ticket: https://trello.com/c/VLDK9GUW/196-refreshing-content

Just added a listener to the AppState which returns active when the app is foregrounded.
Have removed the initial call to getEvents in the root componentDidMount to avoid duplication of this on startup, as the AppState initially returns unknown on first mounting, and switches to active shortly after.

Attempting to include a unit test for the handler but running into issues due to BugSnag comme ça: https://github.com/bugsnag/bugsnag-react-native/issues/205 where the suggested mocking is also not working for me. 

## Tester check-list

  * [ ] Galaxy S7 (Android 6.0)
  * [ ] iPhone X (iOS 11.x)
  * [ ] iPhone 8 (11.x)
  * [ ] iPhone 7 (iOS 10.x)

  Optional:

  * [ ] Google Pixel 2
  * [ ] Galaxy S8 Plus
  * [ ] Galaxy S7 Edge
  * [ ] Galaxy S6
  * [ ] iPhone 7+
  * [ ] iPhone 6

* [ ] Large Text (=<200%)

* [ ] Airplane mode
* [ ] 2G/3G Network Settings


declanslevin:
Is this meant to refresh so that newly added events will appear in the Events listing? 
Because that's not happening currently. I see the `Loading...` message after foregrounding the app, but it still requires a manual pulldown-refresh to get new events to appear.

charypar:
This should now work as expected:

- When the app is started, it will read and use the local cached events, or fetch from the CMS if there is no cache
- Whenever the app is foregrounded, it will quietly check for the CMS data and if anything new arrives, update the content it's showing to the user. This _might_ be somewhat jarring, so I've added support for adding some UI later on that will notify the user of new content being available and give them ability to explicitly display it.

PR #323, opened 2018-05-18
Nightly build which runs only once for a commit

frigus02:
## Pre-flight check-list

Before raising a pull request

* [ ] Documentation
* [ ] Unit tests

## Pre-merge check-list

* [ ] Test plan provided
* [ ] Tester approved

## Tester check-list

  * [ ] Galaxy S7 (Android 6.0)
  * [ ] iPhone X (iOS 11.x)
  * [ ] iPhone 8 (11.x)
  * [ ] iPhone 7 (iOS 10.x)

  Optional:

  * [ ] Google Pixel 2
  * [ ] Galaxy S8 Plus
  * [ ] Galaxy S7 Edge
  * [ ] Galaxy S6
  * [ ] iPhone 7+
  * [ ] iPhone 6

* [ ] Large Text (=<200%)

* [ ] Airplane mode
* [ ] 2G/3G Network Settings


frigus02:
@markholland The nightly build started at 8am now (shifted by 1 hour for some reason). What do you think? Merge like this?

.circleci/config.yml
```diff
@@ -10,6 +10,8 @@ gems-key: &gems-key
   v1-gems-{{ checksum "Gemfile.lock" }}
 brew-key: &brew-key
   v1-brew-{{ checksum "BrewFile" }}
+nightly-key: &nightly-key
+  v0.1-nightly-{{ checksum "~/.pride-nightly/last-commit" }}
```
markholland: I don't think it's a problem but I think two caches would be generated, one for docker and one for OSX because ~/ expands to a different path in each.

.circleci/config.yml
```diff
@@ -479,6 +490,34 @@ jobs:
...
           path: /tmp/screenshots
 
+  test-nightly:
+    <<: *job-defaults-macos
+    working_directory: /Users/distiller/project/ios
+    environment:
+      <<: *env-macos
+    steps:
+      - <<: *checkout-macos
+      - run:
```
markholland: Maybe extract this and below to an anchor now as it will be needed in each release job anyways? I found it easy to make typos when doing this type of refactor, it would also reduce the changes needed to use in release jobs.

markholland:
Let's get this in so that we can get release builds going on Friday

frigus02:
I agree with your comments but I couldn't find the time to work on it recently. I don't mind merging now and fixing this later, though.

I think if we merge this we need to change the branch filter for the nightly builds to "master".

PR #324, opened 2018-05-18
Switch to more deterministic date/time library

### LINKED ISSUE #247: Midnight events appear on wrong day when device is in different timezone
## Description of issue

Events occurring at midnight appear in wrong day in event list when device is in a different timezone.

This seems to be an issue with the `date-fns` library applying timezones to processed dates when the `format` function is called. (Similar behaviour has been observed in other functions too).

## Expected behaviour

Events should always show the date and time of the event as it will be in London on the day. Essentially no timezone transformations should be applied.

## Steps to reproduce

Change your device timezone to Toronto, Canada (UTC-4). If using the iOS simulator you will need to do this on your mac system settings.

See that "Midnight recurring dates test event" which occurs at 00:00 on 08 Aug, displays on the event list as 07 Aug

Tested in iOS simulator on iPhone 6

## Screenshots

screen shot 2018-05-07 at 12 09 41

screen shot 2018-05-07 at 12 09 49
screen shot 2018-05-07 at 12 09 43

### LINKED ISSUE #248: Issues with date-fns library
I've found a few issues with the date-fns library applying default ES timezones when some functions are called.

I looks like many of its functions apply the current device timezone to dates without you asking. We had seen this before in our unit tests but I hadn’t quite appreciated the extent of the problem.

For example if I set my machine time to UTC-4 and re-run the tests or simulator then this happens:

```javascript
const dateTime = '2018-05-06T00:00+00:00';

console.log(formatDate(dateTime)) // 2018-05-05T20:00:00.000-04:00
console.log(getDate(dateTime)) // 5
```

You could half expect this for `formatDate` but it seems to do the same thing for functions like `addDays`, so if you want to do maths on dates you’re gonna have a bad time.

Looks like the library authors are aware of this and are working on a huge new set of functions that will just use UTC: https://github.com/date-fns/date-fns/issues/376

I’ve found a bug in our app due to this. With the same timezone set (UTC-4), an event starting at midnight is grouped into the previous day on the event list. #247

In `moment.js` I don’t remember this being a problem, you simply did the maths on a date and apply the timezone after when you want to display it.

I'm starting to come to the conclusion that `date-fns` isn't really fit for our purposes, as it's important we always display times as they will be on the date and location of an event, not for the user's device in their current location.

To avoid possible undetected and future bugs, it may well be worth switching to another library like `moment.js` or `luxon` which allow you to work with dates irrespective of timezone if you need to.

robbiemccorkell:
Fixes #248
Fixes #247 

This PR switches our date/time library from date-fns to [luxon](https://moment.github.io/luxon/index.html)

A few of us had found that date-fns can be quite unpredictable when handling timezones, as described in #248.

Luxon is more timezone aware, but crucially has the option to base operations around the timezone specified in an ISO string, and not modify it.

Design decisions:
- All interactions with the date/time library have been moved to `src/lib/date` and wrapped in a thin API. (This helped a lot to safely change the functions one by one from date-fns to luxon).
- The data primitive in our app is ISO strings. All interactions with the date lib functions take ISO strings in, and return strings. This helps to avoid unpredictable mixes of data types at different levels of the application.
- All dates are parsed with the luxon `{ setZone: true }` option applied. This forces luxon to respect the timezone specified in the date string and not change it.
- Dates are returned from the lib in a format similar to contentful for safety (in case there is still any code that relies on that format). This format is an ISO string excluding milliseconds and seconds.
- Unit tests always run in the UTC timezone to maintain consistency between local envs and CI. It can also help highlight timezone specific issues early.

## Pre-flight check-list

Before raising a pull request

* [x] Documentation
* [x] Unit tests

## Pre-merge check-list

* [ ] Test plan provided
* [ ] Tester approved

## Tester check-list

  * [ ] Galaxy S7 (Android 6.0)
  * [ ] iPhone X (iOS 11.x)
  * [ ] iPhone 8 (11.x)
  * [ ] iPhone 7 (iOS 10.x)

  Optional:

  * [ ] Google Pixel 2
  * [ ] Galaxy S8 Plus
  * [ ] Galaxy S7 Edge
  * [ ] Galaxy S6
  * [ ] iPhone 7+
  * [ ] iPhone 6

* [ ] Large Text (=<200%)

* [ ] Airplane mode
* [ ] 2G/3G Network Settings

### DIFF
--- a/src/components/CategoriesPills.js
+++ b/src/components/CategoriesPills.js
@@ -21,6 +21,14 @@ type Props = {
   onPress?: Function
 };
 
+const createAccessibilityLabel = (
+  listPrefix: string,
+  pillsList: Set<EventCategoryName>
+) =>
+  pillsList.size > 0
+    ? `${listPrefix} ${[...pillsList].join(", ")}`
+    : text.categoryFilterEmpty;
+
 class CategoriesPills extends React.PureComponent<Props> {
   static defaultProps = {
     style: {}
@@ -55,13 +63,23 @@ class CategoriesPills extends React.PureComponent<Props> {
     const { style, selectedCategories } = this.props;
 
     return (
-      <View style={[styles.selectedCategoriesPills, style]}>
+      <View
+        style={[styles.selectedCategoriesPills, style]}
+        accessible
+        accessibilityLabel={createAccessibilityLabel(
+          text.categoryFilterContents,
+          selectedCategories
+        )}
+      >
         {selectedCategories.size === 0 ? (
           <Text type="h3" style={styles.zeroSelected}>
             {text.zeroSelected}
           </Text>
         ) : (
-          <View style={styles.scrollView}>
+          <View
+            style={styles.scrollView}
+            importantForAccessibility="no-hide-descendants"
+          >
             <LinearGradient
               style={[styles.scrollShadow, styles.scrollShadowLeft]}
               end={{ x: 0, y: 0.5 }}
--- a/src/components/CheckBox.js
+++ b/src/components/CheckBox.js
@@ -4,6 +4,7 @@ import { Image, StyleSheet } from "react-native";
 import type { ViewStyleProp } from "react-native/Libraries/StyleSheet/StyleSheet";
 import Text from "./Text";
 import Touchable from "./Touchable";
+import { checkboxAccessibilityLabel } from "../data/accessibility";
 
 import checkboxUrl from "../../assets/images/checkBox.png";
 import checkBoxCheckedUrl from "../../assets/images/checkBoxChecked.png";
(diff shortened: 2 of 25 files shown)


src/selectors/events.js
```diff
@@ -144,32 +124,12 @@ export const groupPerformancesByPeriod = (
...
-          return { periods: [...periods, buffer], buffer: [performance] };
-
-        return { periods, buffer: [...buffer, performance] };
-      },
-      { periods: [], buffer: [] }
-    );
-
-  return sections.buffer.length > 0
-    ? [...sections.periods, sections.buffer]
-    : sections.periods;
```
robbiemccorkell: Refactoring these two large functions might seem unrelated, but it allowed me to remove the need for `parseDate`, and `differenceInCalendarDays` functions. These were perfectly possible in `luxon`, but I saw an opportunity to simplify.

declanslevin:
I'm not entirely sure how to test this properly - I've tried changing timezones on my phone but that's not affecting how events appear. Playing around with changing start times doesn't seem to do much either. 

@robbiemccorkell when you get a chance, could you provide some guidance on what I can do to test this effectively?

robbiemccorkell:
@declanslevin Well that's a good sign. I have a way to test #247. I added an event to the test contentful space called "Midnight Event". It's simply an event with a start time of 00:00 on a particular day, nothing special.

Currently in the app, if the device timezone is set to something behind UTC (I tested in Toronto, Canada which is UTC-4 I think), you should see that the event appears in the event list a day before it actually occurs.

If you do the same test off this branch, you should see that the event appears on the expected day group in the event list.

PR #328, opened 2018-05-22
Hide past events

RGBboy:
**[Whole app - Not showing Past Events](https://trello.com/c/JVwbl10D/154-whole-app-not-showing-past-events)**

This PR hides past events in the event listing, featured events and saved events screens. I have made it so we update a piece of state in the event filters reducer called `showEventsAfter` when we navigate to certain screens (as outlined in the trello ticket). Happy to chat about this approach.

Also noteworthy is the `showEventsAfter` state is a `luxon DateTime`. This is inconsistent with how we store other dates as strings however gives the benefit of begin able to type the thing so we know it is a date.

A follow up PR will be made to disabled past dates in the calendar filter component.

## Tester check-list

  * [ ] Galaxy S7 (Android 6.0)
  * [ ] iPhone X (iOS 11.x)
  * [ ] iPhone 8 (11.x)
  * [ ] iPhone 7 (iOS 10.x)


src/reducers/event-filters.js
```diff
@@ -1,9 +1,18 @@
...
 import type { State } from "../data/event-filters";
+import type { NavigationAction } from "../actions/navigation";
+import type { InitAction } from "../actions";
+import { NAVIGATION } from "../actions/navigation";
+import { HOME, PARADE, SUPPORT_US, DONATE, SPONSOR } from "../constants/routes";
 
-const defaultState = {
+type Action = EventFiltersAction | NavigationAction | InitAction;
+
+const includedRoutes = [HOME, PARADE, DONATE, SPONSOR, SUPPORT_US];
```
markholland: Should this be defined somewhere close to our navigation so we don't have to remember to update this if tabs change?

RGBboy: Sure, I'll move it to the routes constants.

src/constants/routes.js
```diff
@@ -5,7 +5,6 @@ export const SAVED_EVENT_LIST = "SAVED_EVENT_LIST";
 export const HOME = "HOME";
 export const EVENT_CATEGORIES_FILTER = "EVENT_CATEGORIES_FILTER";
 export const PARADE = "PARADE";
```
markholland: Shouldn't `SAVED_EVENT_LIST` be here?

src/selectors/events.js
```diff
@@ -199,15 +209,17 @@ export const selectFeaturedEventsByTitle = (state: State, title: string) => {
     return [];
   }
 
-  return ((featured.fields.events[locale].map(e =>
-    selectEventById(state, e.sys.id)
-  ): any): Event[]);
+  return featured.fields.events[locale].reduce((acc, item: Reference) => {
+    const event: ?Event = selectEventById(state, item.sys.id);
+    if (event) {
+      acc.push(event);
```
markholland: Wouldn't `return acc.concat(event)` be preferrable?

RGBboy: I would normally say yes, but because this reducer is in the same block as the initial, we can reuse the underlying array. Happy to change, just didn't see the reason to allocate more arrays.

markholland: Makes sense, good to leave

RGBboy:
@robbiemccorkell would love your input on my use of Luxon.

> Also noteworthy is the `showEventsAfter` state is a luxon `DateTime`. This is inconsistent with how we store other dates as strings however gives the benefit of begin able to type the thing so we know it is a date.

robbiemccorkell:
I think it makes sense to me. Was going to comment on that we aren't using a string as date primitive, but just saw your comment above. I guess the issue I was trying to avoid, of not knowing exactly what a date var is if we pass around different things isn't so much of a problem when we have a type system!

The only thing I'd mention is to be careful when using `fromISO`. The few places you're using it here I think look fine, but in some cases you need `setZone:  true` to get expected behaviour regarding the display of a parsed ISO string. 

To hide away this I tried to design the `lib/date` module to always accept ISO strings for each function, and to parse them in a safe way behind the scenes so as a dev you don't have to worry about it.

For this code we could for example add 2 new functions into the `lib/date` module `now: void => string` and `isAfter: (string, string) => bool`. The reducer could call `now` directly rather than needing it passed into scope, and the selectors could `isAfter` as a generic function without having to worry about whether the parsing logic is correct.

Just some thoughts, nothing essential. I think the code should work fine as is.

PR #344, opened 2018-05-30
Refactor events to data

RGBboy:
PR to refactor event actions and reducer to be called data. Originally we were only getting events from the CMS, however now we get a number of different types of content. This PR just tries to make that a bit clearer.


src/actions/data.js
```diff
@@ -2,19 +2,19 @@
...
 import { getCmsData, updateCmsData } from "../integrations/cms";
 import type { SavedData } from "../integrations/cms";
-import type { StandardAction } from "./";
 
-type CmsActionType =
-  | "REQUEST_CMS_DATA"
-  | "RECEIVE_CMS_DATA"
-  | "REQUEST_UPDATE_CMS_DATA";
-
-export type CmsAction = StandardAction<CmsActionType, SavedData>;
```
RGBboy: Instead of using the StandardAction, which makes the payload optional, I have explicitly stated the sape of each kind of action. This allows us to access deeper into the data property with checking if it exists.

PR #345, opened 2018-05-31
Sponsor refactor

RGBboy:
This PR is the first of a series I would like to make where we refactor towards safely decoding the CMS data into a completely type checked data model. Of note here is the use of a decoder to decode a Sponsor type into the Sponsor array in the data reducer.

Decoders allow us to compose transformation functions. These functions take in a `mixed` type and output a Result which is either ok, and has the concrete type on the value property, or is in an error state with an error message on the error property. In the case of the sponsor type:

```js
type SponsorDecoder = mixed => Result<string, Sponsor>
```

This encourages us to handle failure when decoding, in the sponsors case, defaulting to an empty array. (see `reducers/data.js`)

I have also taken this opportunity to simplify how we handle locale in the sponsor model. Having the type `{ [string]: Field }` is prone to error, as the property can be any string and even be missing. Instead, we hoist the locale during decoding so we no longer need to cater for it being missing.

## Tester check-list

  * [ ] Galaxy S7 (Android 6.0)
  * [ ] iPhone X (iOS 11.x)
  * [ ] iPhone 8 (11.x)
  * [ ] iPhone 7 (iOS 10.x)

### DIFF
--- a/src/lib/result.test.js
+++ b/src/lib/result.test.js
@@ -1,0 +1,52 @@
+// @flow
+import { ok, error, map, withDefault } from "./result";
+
+describe("ok", () => {
+  it("creates an ok result", () => {
+    const result = ok("test");
+    expect(result.ok).toEqual(true);
+    if (result.ok) {
+      expect(result.value).toEqual("test");
+    }
+  });
+});
+
+describe("error", () => {
+  it("creates an errored result", () => {
+    const result = error("message");
+    expect(result.ok).toEqual(false);
+    if (!result.ok) {
+      expect(result.error).toEqual("message");
+    }
+  });
+});
+
+describe("map", () => {
+  it("maps ok results", () => {
+    const result = map(value => value + 1, ok(1));
+    expect(result.ok).toEqual(true);
+    if (result.ok) {
+      expect(result.value).toEqual(2);
+    }
+  });
+
+  it("does not map errored results", () => {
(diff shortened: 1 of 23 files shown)


src/reducers/data.js
```diff
@@ -1,30 +1,41 @@
 // @flow
-import type { Reducer } from "redux";
 import type { DataAction } from "../actions/data";
 import type { CmsEntry } from "../integrations/cms";
 import type { Asset } from "../data/asset";
+import type { Sponsor } from "../data/sponsor";
+import decodeSponsor from "../data/sponsor";
+import locale from "../data/locale";
+import type { Decoder } from "../lib/decode";
+import { filterMap as decodeFilerMap } from "../lib/decode";
```
frigus02: 🔍 decodeFilerMap: missing "t"

src/lib/result.test.js
```diff
@@ -0,0 +1,52 @@
+// @flow
+import { ok, error, map, withDefault } from "./result";
+
+describe("ok", () => {
+  it("creates an ok result", () => {
+    const result = ok("test");
+    expect(result.ok).toEqual(true);
+    if (result.ok) {
```
frigus02: This this `if` necessary because of flow?

RGBboy: yeah, a bit annoying because I had to wrap all the Result value/error assertions with this.

PR #368, opened 2018-06-02
Fix crash on Android when animating too many items

frigus02:
I noticed the app crashes on Android when you start to filter events. This was introduced in #358. It seems to crash when animating too many items at once.

To fix this, I made it animate only when 1 event was added/removed from an list.

Also I noticed a bug in our render optimization. We wanted to prevent rendering tab views when they are not currently visible and did this by using `navigation.isFocused()` in `shouldComponentUpdate`. This didn't work for me on Android. But there is a nicer solution anyway using `withNavigationFocus`. I had to change the flow types a bit. Would be great if you could have a look at that, @RGBboy.

## Pre-flight check-list

Before raising a pull request

* [ ] Documentation
* [ ] Unit tests

## Pre-merge check-list

* [ ] Test plan provided
* [ ] Tester approved

## Tester check-list

  * [ ] Galaxy S7 (Android 6.0)
  * [ ] iPhone X (iOS 11.x)
  * [ ] iPhone 8 (11.x)
  * [ ] iPhone 7 (iOS 10.x)

  Optional:

  * [ ] Google Pixel 2
  * [ ] Galaxy S8 Plus
  * [ ] Galaxy S7 Edge
  * [ ] Galaxy S6
  * [ ] iPhone 7+
  * [ ] iPhone 6

* [ ] Large Text (=<200%)

* [ ] Airplane mode
* [ ] 2G/3G Network Settings


src/screens/EventsScreen/component.js
```diff
@@ -30,13 +30,15 @@ export type Props = {
...
+  selectedCategories: Set<EventCategoryName>
 };
 
-class EventsScreen extends Component<Props> {
-  shouldComponentUpdate = () => this.props.navigation.isFocused();
+type AllProps = Props & NavigationProps;
+
+class EventsScreen extends Component<AllProps> {
+  shouldComponentUpdate(nextProps: AllProps) {
+    return nextProps.isFocused;
```
charypar: Neat 📸

PR #369, opened 2018-06-02
Spike: Alert user to new version in app store

robbiemccorkell:
This PR uses the package [`react-native-version-check`](https://github.com/kimxogus/react-native-version-check) to check if there is a newer version of the app in a respective app store and alerts the user.

The package seems to compare the latest app store version to that found in `Info.plist` in IOS, and the `versionName` defined in Android. I'm not very familiar with how this works but I assume we set these numbers via fastlane at build time.

To manually test this on each platform we had to:

IOS:
- Change the version number in `Info.plist` to something lower than that listed in the app store.

Android:
- Change the `run-android` command in `package.json` to `org.prideinlondon.festival` instead of `org.prideinlondon.festival.dev` to match that app store app Id.
- In `android/app/build.gradle` change `productFlavors.dev.applicationId` to `org.prideinlondon.festival`.

I'm sure there's a better way of doing this, but those are the steps we took.

If everyone thinks this is feasible we can run it past UX to see what the best way of alerting the user is, and to consider a "Remind me later" feature.

## Pre-flight check-list

Before raising a pull request

* [ ] Documentation
* [ ] Unit tests

## Pre-merge check-list

* [ ] Test plan provided
* [ ] Tester approved

## Tester check-list

  * [ ] Galaxy S7 (Android 6.0)
  * [ ] iPhone X (iOS 11.x)
  * [ ] iPhone 8 (11.x)
  * [ ] iPhone 7 (iOS 10.x)

  Optional:

  * [ ] Google Pixel 2
  * [ ] Galaxy S8 Plus
  * [ ] Galaxy S7 Edge
  * [ ] Galaxy S6
  * [ ] iPhone 7+
  * [ ] iPhone 6

* [ ] Large Text (=<200%)

* [ ] Airplane mode
* [ ] 2G/3G Network Settings

### DIFF
--- a/android/app/build.gradle
+++ b/android/app/build.gradle
@@ -223,6 +223,7 @@ repositories {
 }
 
 dependencies {
+    implementation project(':react-native-version-check')
     implementation project(':react-native-haptic-feedback')
     implementation project(':lottie-react-native')
     implementation project(':react-native-splash-screen')
--- a/android/app/src/main/java/org/prideinlondon/festival/MainApplication.java
+++ b/android/app/src/main/java/org/prideinlondon/festival/MainApplication.java
@@ -16,6 +16,7 @@ import com.facebook.react.ReactNativeHost;
 import com.facebook.react.ReactPackage;
 import com.facebook.react.shell.MainReactPackage;
 import com.facebook.soloader.SoLoader;
+import io.xogus.reactnative.versioncheck.RNVersionCheckPackage;
 
 import io.fabric.sdk.android.Fabric;
 import java.util.Arrays;
@@ -40,7 +41,8 @@ public class MainApplication extends Application implements ReactApplication {
             new MapsPackage(),
             new LinearGradientPackage(),
             new ReactNativeConfigPackage(),
-            new AnalyticsPackage()
+            new AnalyticsPackage(),
+            new RNVersionCheckPackage()
       );
     }
 
--- a/android/settings.gradle
+++ b/android/settings.gradle
@@ -13,5 +13,7 @@ include ':react-native-config'
 project(':react-native-config').projectDir = new File(rootProject.projectDir, '../node_modules/react-native-config/android')
 include ':react-native-linear-gradient'
 project(':react-native-linear-gradient').projectDir = new File(rootProject.projectDir, '../node_modules/react-native-linear-gradient/android')
+include ':react-native-version-check'
+project(':react-native-version-check').projectDir = new File(rootProject.projectDir, 	'../node_modules/react-native-version-check/android')
 
 include ':app'
--- a/ios/PrideLondonApp.xcodeproj/project.pbxproj
+++ b/ios/PrideLondonApp.xcodeproj/project.pbxproj
@@ -51,6 +51,7 @@
 		C7BF4CAE82A24794807F8B14 /* Roboto-Bold.ttf in Resources */ = {isa = PBXBuildFile; fileRef = 39B45C6580BE4CD4A6D4DEBC /* Roboto-Bold.ttf */; };
 		CEAF234A6DB24F138A0FDE42 /* Poppins-SemiBoldItalic.ttf in Resources */ = {isa = PBXBuildFile; fileRef = 0551B12566AC4BE3876EE344 /* Poppins-SemiBoldItalic.ttf */; };
 		D4D47B1F77B94F2F8818448B /* libReactNativeConfig.a in Frameworks */ = {isa = PBXBuildFile; fileRef = 6D3680107BE344DF99D16CDA /* libReactNativeConfig.a */; };
+		D9973E5120C2C5260007257F /* libRNVersionCheck.a in Frameworks */ = {isa = PBXBuildFile; fileRef = D9973E4F20C2C4D20007257F /* libRNVersionCheck.a */; };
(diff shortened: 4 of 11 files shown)


lani-s:
@clmntnbrn @stevecottle would be good to get some UXD input into how we might surface this to users without annoying them :)

robbiemccorkell:
@markholland When you next have the chance it would be great to get your opinion on this. Want to make sure that it fits with how we are managing version numbers.

robbiemccorkell:
I've added "Ask me later" functionality and productionised the code. In it's current state this could go out to production if we really want it and there's always the option to make it prettier later.

If a user presses "Ask me later" the app will wait until the start of the next day to notify the user when they next open the app.

screen shot 2018-06-09 at 17 24 48
screen shot 2018-06-09 at 17 24 49

robbiemccorkell:
Could definitely do with a review of the copy in the dialog too

src/components/StoreVersionDialog.js
```diff
@@ -0,0 +1,86 @@
...
+      !!askLaterTime && isBefore(now(), askLaterTime);
+
+    const showAppStoreAlert = versionInfo.isNeeded && !updateMessageSuppressed;
+    if (showAppStoreAlert) {
+      this.handleAppStoreAlert();
+    }
+  }
+
+  handleAppStoreAlert = () =>
+    Alert.alert(
```
frigus02: Will this work if the app was send to the backround while the update check was running?

src/components/StoreVersionDialog.test.js
```diff
@@ -0,0 +1,114 @@
...
+import { shallow } from "enzyme";
+import { Alert } from "react-native";
+import VersionCheck from "react-native-version-check";
+import configureStore from "redux-mock-store";
+import thunk from "redux-thunk";
+import StoreVersionDialog from "./StoreVersionDialog";
+import * as storage from "../integrations/storage";
+import * as date from "../lib/date";
+
+const nextTick = () => new Promise(resolve => setTimeout(resolve, 0));
```
frigus02: Do you think we can use [`jest.runAllTicks()`](https://facebook.github.io/jest/docs/en/jest-object.html#jestrunallticks) instead?

robbiemccorkell: I didn't know about this, but I tried it out and it gave me a weird warning and didn't work.

I've simplified it a little to use `process.nextTick` though, since I only just found out about this too.

src/components/StoreVersionDialog.js
robbiemccorkell: I did a quick test by throwing a delay into the code. It looks like when the app is in the background, nothing happens on the system level, but when you open the app again the dialog is there.

frigus02: Nice, that sounds good :+1:

src/actions/reporting.js
```diff
@@ -0,0 +1,11 @@
+// @flow
+import type { Client } from "bugsnag-react-native";
+import type { Dispatch } from "redux";
+import type { State } from "../reducers";
+
+// eslint-disable-next-line import/prefer-default-export
+export const notifyReporter = (error: Error) => (
+  _: Dispatch<>,
+  __: () => State,
+  errorReporter: Client
```
RGBboy: I wonder if it would be better to create a top level action, like our INIT action that is just called ERROR with an error property to hold the error data. Then to notify bugsnag, we create a bugsnag middleware that listens for this action. This would mean that we don't need to instantiate the thunk middleware with extra arguments and complicate our action creators.

robbiemccorkell: Sure thing, makes sense. Sorry it took me so long to get back to this, I'll fix this up on Saturday

markholland:
Is it possible to only show this for minor updates rather than patches? Because of our current release process, if there's a bug on one platform we still have to release a patch update for both.

robbiemccorkell:
@markholland Yea I think you can specify a depth to the library. That's a good point, it might be a bit too chatty on on patch versions.

I'll have a go at limiting to major and minor only.

stevecottle:
This looks fine, I have a couple of questions about the details:

How often will this be triggered?
What happens when I tap 'OK'?

Could the copy read:
Update app?
A new version of this app is available from the store. Would you like to update now?
[ Ask me tomorrow ]                       [ Disagree ]    [ Agree ]

Ref:
https://material.io/design/components/dialogs.html#anatomy 

I can't find an iOS version of this doc^ for guidance of button order. Has anyone copme across anything similar?

serdemli:
@stevecottle 
The iOS guide on Alerts is here: https://developer.apple.com/design/human-interface-guidelines/ios/views/alerts/

We should be able to change the CTA text- and I agree it should be clearer than just "OK" - "Update now" sounds good to me. That CTA would take you to the app store or google play, where you can update the app. 

"Ask me tomorrow" could be a weird one, as from what I understand, you can customise when you want to get reminders on Android. Ask me later is the generic and most common way most apps go about it. Rather than having to build something different for Android- if that's the case - we should keep both the same and keep it consistent with other apps out there.

So.. @robbiemccorkell sounds like this one's good to go, with the change from OK to Update now, if possible :)

frigus02:
I can have a look at changing the text today and limit to major and minor versions only.

frigus02:
@stevecottle: It will be triggered on every app start, but not when the app is already running and just brought to the foreground.

@serdemli Android won't be an issue. Android restricts usage of exact timers and users can choose to disable/deley notifications. But we use neither of those in this PR.

frigus02:
We discussed in stand-up, that this might not be needed anymore since won't get the notification for the parade update anyway. So I'll let @robbiemccorkell finish this 🙂

PR #370, opened 2018-06-02
Refactor test data to use testcheck generators

RGBboy:
So we have a lot of test data all over the place. This PR starts to migrate us towards using deterministic generators to create test data for various types. As I continue to refactor our core data types I will convert more of our tests towards this way.

This is part of a bigger piece of eventually enabling type checking in our tests. Currently we can not do this because in a number of places we bend our type rules.


src/data/__test-data.js
```diff
@@ -0,0 +1,143 @@
+// @flow
+// eslint-disable-next-line import/no-extraneous-dependencies
```
frigus02: Is that ESLint exception needed?

RGBboy: Yeah, ESLint complains about it being a devDependency. I'm guessing it thinks this file is application code.

frigus02: Ah right. I see we already ignore other files here. So maybe we should do the same for this file?

https://github.com/redbadger/pride-london-app/blob/9ac60fe91e9239e40fd3a0fb027d454f69089498/.eslintrc#L40-L47

PR #371, opened 2018-06-03
Optimise event screen update

RGBboy:
So I think I found the root cause of our performance problem when navigating between tabs. Essentially the EventList connector was getting called every time, and therefore recomputing the event list state (filtering etc). This was fixed by removing the route reducer (which was never used but would trigger this on every route change) and also only updating showEventsAfter after 30 minutes since it last updated.

I have only tested in iOS simulator. It would be good to try out the alpha build of this before we merge just to confirm.


markholland:
On iOS for me it’s only faster if I’ve visited that tab before.

RGBboy:
Yeah, essentially as soon as any state changes the EventList has the mapStateToProps run. I'm hoping my other refactoring that I am working on will help mitigate this too. There might be some other things we can do to mitigate this. Perhaps connecting at the event card level instead for state like saved etc. I'll see if there are any other changes I can make and will submit more PR's if I find anything.

PR #375, opened 2018-06-04
Asset refactor

RGBboy:
This PR applies the same pattern as established in #345 to the Assets (well just images, because that is the only asset we use). `state.data` now contains a dictionary of imageDetails that are type safe.

This doesn't completely hook images up with all our components, but instead tries it out with the SponsorScreen. Of note is the fact that we may not have details for an image, as documented in the getImageDetails function:

```js
export const getImageDetails = (images: Images) => (
  id: string
): ?ImageDetails => images[id];
```

At some point I think it would be a good idea to create our own Image component that falls back when these details are missing (could just have a grey background or similar).

Once other components are refactored, we can remove the assets array in state and remove the getAssetSource functionality.

Keen to hear any thoughts/objections.

### DIFF
--- a/src/data/__snapshots__/performance.test.js.snap
+++ b/src/data/__snapshots__/performance.test.js.snap
@@ -1,6 +1,6 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
-exports[`Performance decoder correctly decodes valid CMS sponsor 1`] = `
+exports[`Performance decoder correctly decodes valid CMS performance 1`] = `
 Object {
   "contentType": "performance",
   "fields": Object {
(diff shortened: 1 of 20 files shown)


src/data/__snapshots__/performance.test.js.snap
```diff
@@ -1,6 +1,6 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
-exports[`Performance decoder correctly decodes valid CMS sponsor 1`] = `
+exports[`Performance decoder correctly decodes valid CMS performance 1`] = `
```
RGBboy: missed this one in an earlier PR :(

src/data/performance.test.js
```diff
@@ -4,7 +4,7 @@ import decodePerformance from "./performance";
 
 describe("Performance", () => {
   describe("decoder", () => {
-    it("correctly decodes valid CMS sponsor", () => {
+    it("correctly decodes valid CMS performance", () => {
```
RGBboy: Missed this one in an earlier PR :(

PR #376, opened 2018-06-04
Event data refactor

RGBboy:
This PR applies the same pattern as established in #345 to the Events. state.data now contains an array of Events that are type safe. This PR does not hook it up with our components yet. Once this is in I will do that.

The number of files changed is not as bad as it looks. Most of them are a change of the exiting data/events -> data/events-deprecated.


src/data/__test-data.js
```diff
@@ -114,6 +115,39 @@ export const generateCMSHeaderBanner: ValueGenerator<mixed> = gen({
...
+    eventDescription: "eventDescription",
+    accessibilityDetails: "accessibilityDetails",
+    email: "email",
+    phone: "phone",
+    ticketingUrl: "ticketingUrl",
+    venueDetails: ["venueDetailsA", "venueDetailsB"],
+    individualEventPicture: generateFieldRef,
+    eventsListPicture: generateFieldRef,
+    performances: [],
+    recurrenceDates: []
```
frigus02: Do you think we should generate performances and recurrence dates? Or is this never used in tests?

src/data/event.js
```diff
@@ -98,10 +32,212 @@ export type EventCategoryName =
...
+
+  const difference = diffDate(recurrenceStartTime, startTime);
+  const recurrenceEndTime = addToDate(endTime, difference);
+
+  return R.mergeDeepRight(event, {
+    fields: {
+      startTime: formatDate(recurrenceStartTime, FORMAT_CONTENTFUL_ISO),
+      endTime: formatDate(recurrenceEndTime, FORMAT_CONTENTFUL_ISO),
+      recurrenceDates: [
+        formatDate(startTime, FORMAT_EUROPEAN_DATE),
```
frigus02: Could use `formatEuropeanDate` or maybe remove `formatEuropeanDate`?

src/data/__test-data.js
RGBboy: I've left these empty for now, but I have a task to fill in some of the blanks once we get the base of all these changes in. One thing I would like to do is refactor some of our tests to use the generators too. I think at that stage, filling these in becomes beneficial.

PR #380, opened 2018-06-05
Play Store content changes

markholland:
https://trello.com/c/gh4k1TDW/173-content-for-play-store-listing


markholland:
I'm going to split release notes into a different task because of how it works on Android, we'd have to  know the versionCode that we will be releasing as the file name is versionName.txt

Issue #374, opened 2018-06-04
E2E tests for all screens


RGBboy:
For regression it would be great if we can add tests for our main screens in the app. Currently we only have tests for navigating to the first level tab.

It may be a good idea to try some user flows along with testing getting to each screen. E.g. navigate to an event, refresh content, applying filters, save an event etc.

**Screens:**

- Home - **Tested**
  - Featured Events
- Events - **Tested**
  - Category Filter
  - Date Filter
  - Filter? Filter (area, price etc)
  - Event Details
- Parade - **Tested**
- Saved - **Tested**
- Support - **Tested**
  - Donate
  - Sponsor (Partner)

E2E tests are written here: https://github.com/redbadger/pride-london-app/blob/master/e2e/firstTest.spec.js

markholland:
Also feel free to restructure in any way. Would be good to get at least one example of each type of test we'd like then we can fill out as we have time.

RGBboy:
Related to #302

RGBboy:
Closing this as we now have tests for navigating to all screens. We are still missing tests for filtering etc, but can add these as part of #302

PR #382, opened 2018-06-05
E2E tests

nicollecastrog:
## Pre-flight check-list

Before raising a pull request

* [ ] Documentation
* [ ] Unit tests

## Pre-merge check-list

   Related to issue #374
* [ ] Test plan provided
* [ ] Tester approved

## Tester check-list

  * [ ] Galaxy S7 (Android 6.0)
  * [ ] iPhone X (iOS 11.x)
  * [ ] iPhone 8 (11.x)
  * [ ] iPhone 7 (iOS 10.x)

  Optional:

  * [ ] Google Pixel 2
  * [ ] Galaxy S8 Plus
  * [ ] Galaxy S7 Edge
  * [ ] Galaxy S6
  * [ ] iPhone 7+
  * [ ] iPhone 6

* [ ] Large Text (=<200%)

* [ ] Airplane mode
* [ ] 2G/3G Network Settings


e2e/events.spec.js
```diff
@@ -0,0 +1,86 @@
+const { takeScreenshot } = require("./helpers");
+const { formatDateRange } = require("../src/data/formatters");
+
+const formattedDateNumber = num => (num < 10 ? `0${num}` : num);
+
+const today = new Date();
```
nicollecastrog: I don't love this date logic here, but I can't work out a better way to test the filtering that selecting today through the end of the month... any ideas?

src/components/EventList.js
```diff
@@ -32,21 +33,23 @@ type State = {
   eventsReordered: boolean
 };
 
+type Section = SectionBase<Event> & { index: number };
+
 type RenderItemInfo = {
-  item: Event // eslint-disable-line react/no-unused-prop-types
+  item: Event, // eslint-disable-line react/no-unused-prop-types
+  index: number,
+  section: Section
```
nicollecastrog: the `renderItem` function provides `index` and `section` as subsequent args

src/components/EventList.js
```diff
@@ -132,6 +136,7 @@ class EventList extends Component<Props, State> {
           addSavedEvent={addSavedEvent}
           removeSavedEvent={removeSavedEvent}
           onPress={onPress}
+          testID={`event-card-${index}-${section.index}`}
```
nicollecastrog: I don't love making the `testID` *less* unique by using the two indices (rather than the event ID). However, for the e2e tests, unlike the component-level tests, we can't load test data (and therefore know what the event ID will be), the data comes from Contentful itself. So given it's only a `testID` (and not something important like a `key`) I thought it might be ok to use the indices in this manner... Thoughts?

The reason each event card needs a `testID` is so the e2e tests can explicitly pick out the first event and test saving the event as well as tapping through to Event Details. Detox does provide an `.atIndex(n)` that I wanted to use, but there appears to be a bug because `.atIndex(0)` keeps picking random events in the middle of the list, off screen, so I can't test what's needed because the element won't be interactable

e2e/support-us.spec.js
```diff
@@ -0,0 +1,39 @@
...
+  });
+
+  it("should display links to relevant sections", async () => {
+    await element(by.id("support-us-tab-button")).tap();
+
+    await expect(element(by.id("support-us-donate-button"))).toBeVisible();
+    await expect(element(by.id("support-us-volunteer-button"))).toBeVisible();
+    await expect(element(by.id("support-us-merchandise-button"))).toBeVisible();
+
+    await element(by.id("support-us-main-content")).scrollTo("bottom");
```
nicollecastrog: This is the scrolling I can't get working, as referenced in the TODO section of the PR description...

frigus02: I wonder of the scroll does not work because the "tap" happens on the tiny shadow we added on top the tab navigation. This shadow is positioned absolute above the scroll view. My fingers are not small enough to verify this on a physical device but I can reproduce this with the mouse cursor in an emulator. Trying to scroll by tapping on this shadow does not work. 🤔

e2e/events.spec.js
frigus02: I think that should be fine.

e2e/support-us.spec.js
frigus02: Yeah, this is the reason. We fixed it by adding `pointerEvents="none"` to the shadow view.

PR #384, opened 2018-06-05
Asset switcharoo

RGBboy:
Follow up PR to #375 in order to switch the rest of our components to use `state.data.images` to get image details. Of note is the refactor to use connected images to pull the source data from state which greatly simplifies the passing of props.


src/components/Image.js
```diff
@@ -0,0 +1,36 @@
...
+
+export const Image = ({ reference, getImageDetails, ...props }: Props) => (
+  <RNImage source={getImageDetails(reference.sys.id)} {...props} />
+);
+
+// Note we must add a return type here for react-redux connect to work
+// with flow correctly. If not provided is silently fails if types do
+// not line up. See https://github.com/facebook/flow/issues/5343
+const mapStateToProps = (state, ownProps: OwnProps): Props => ({
+  getImageDetails: GetImageDetails(state.data.images),
```
frigus02: I don't like the naming of `GetImageDetails` (with a capital G). I first thought this might be a React component. Maybe we can import it as something like `createImageDetailsGetter`. What do you think?

src/screens/HomeScreen/component.js
```diff
@@ -32,7 +31,7 @@ type Props = {
   featuredEventsTitle: string,
   featuredEvents: Event[],
   loading: boolean,
-  getAssetSource: FieldRef => ImageSource
+  getImageDetails: string => ?ImageDetails
```
RGBboy: I had to keep this here in order to service the HeaderBanners need for dynamic sizing. See src/screens/HomeScreen/Header.js

src/components/Image.js
```diff
@@ -0,0 +1,36 @@
...
+  reference: FieldRef
+};
+
+type StateProps = {
+  getImageDetails: string => ?ImageDetails
+};
+
+type Props = OwnProps & StateProps;
+
+export const Image = ({ reference, getImageDetails, ...props }: Props) => (
```
frigus02: I thought might be nice to type the `...props` as the React Native image props. But I couldn't find flow types for the image component. So we would need to create and maintain our own here 😞. I guess it's fine like it is.

src/components/__snapshots__/Image.test.js.snap
```diff
@@ -0,0 +1,17 @@
...
+exports[`renders correctly 1`] = `
+<Image
+  accessibilityLabel="Test Label"
+  resizeMode="contain"
+  source={
+    Object {
+      "height": 357,
+      "id": "3CU8n6O760GJ80Wr7r",
+      "revision": 1,
+      "uri": "//red-badger.com/G36Nw4N4Qb1HE305M0V.jpg",
```
frigus02: Hm, there might be a bug in `getImageDetails` that I overlooked earlier. Contentful gives us the uri without a scheme. I thought React Native requires a scheme, which is why we added "https:" in `getAssetSource` to the uri before. However I tested the build and it seems to work just fine. I am a bit confused. Does this just work now?!

src/components/Image.js
RGBboy: yeah, I tried looking for that too. Not sure we want to maintain our own copy.

frigus02: Lets see if we miss them when we don't have them here. We can always add them later. 🙂

src/components/__snapshots__/Image.test.js.snap
RGBboy: The code will add the `https` (the logic is in the image decoder). I just forgot to make the test data generator for ImageDetails add this.

src/components/ImageBackground.js
```diff
@@ -0,0 +1,40 @@
+// @flow
+import React from "react";
+import { ImageBackground as RNImageBackground } from "react-native";
+import { connect } from "react-redux";
+import type { Connector } from "react-redux";
+import type { ImageDetails } from "../data/image";
+import { getImageDetails as GetImageDetails } from "../data/image";
```
frigus02: Would you mind renaming `createImageDetailsGetter` it here as well? 😊

PR #385, opened 2018-06-06
Hook up new event model to EventDetails Screen

RGBboy:
Follow up PR to #376 in order to switch the EventDetails screen to use `state.data.events` to get event details. Slowly getting closer to removing the deprecated events and entries.

This PR also adds the initial [reselect](https://github.com/reduxjs/reselect) and starts to refactor our selectors so they can be more easily composed with reselect. Once this pattern is applied to the Event Listing Screen we should see a performance boost as reselect memoizes selectors.

Unfortunately I had to update the way test data is calculated and therefore a number of snapshots updated. I have double checked them but it might be a good idea to get another pair of eyes.


src/screens/EventDetailsScreen/EventOverview.js
```diff
@@ -63,32 +61,30 @@ const EventOverview = ({ event }: Props) => {
...
           {formatLongEventPrice(
-            event.fields.eventPriceLow[locale],
-            event.fields.eventPriceHigh[locale]
+            event.fields.eventPriceLow,
+            event.fields.eventPriceHigh
           )}
         </Text>
       </IconItem>
 
       {event.fields.venueDetails &&
```
frigus02: In the new event type `event.fields.venueDetails` is always an array, so we should be able to remove this check. Similar to `accessibilityOptions`.

src/screens/EventDetailsScreen/component.js
```diff
@@ -19,10 +20,10 @@ import LayoutColumn from "../../components/LayoutColumn";
 import ShadowedScrollView from "../../components/ShadowedScrollView";
 import SectionDivider from "../../components/SectionDivider";
 import PerformanceList from "../../components/PerformanceList";
-import { groupPerformancesByPeriod } from "../../selectors/events";
+// Move this to performance selector
```
frigus02: It is already in the performance selectors, isn't it?

src/screens/EventDetailsScreen/component.js
```diff
@@ -166,20 +163,16 @@ class EventDetailsScreen extends PureComponent<Props> {
...
                       {(event.fields.email || event.fields.phone) && (
                         <EventContact
-                          email={
-                            event.fields.email && event.fields.email[locale]
-                          }
-                          phone={
-                            event.fields.phone && event.fields.phone[locale]
-                          }
+                          email={event.fields.email && event.fields.email}
+                          phone={event.fields.phone && event.fields.phone}
```
frigus02: We should be able to remove these checks and just pass down `event.fields.email` and `event.fields.phone`.

src/screens/EventDetailsScreen/component.js
RGBboy: Yep, originally it wasn't. This was a note for me to do that work that I forgot to remove

src/components/__snapshots__/Image.test.js.snap
```diff
@@ -7,7 +7,7 @@ exports[`renders correctly 1`] = `
   source={
     Object {
       "height": 357,
-      "id": "3CU8n6O760GJ80Wr7r",
+      "id": "4D776SGAHYRnNGHx1DES622EFYZ",
```
markholland: If we update to Jest 23 we can avoid these sorts of snapshot changes where the value is unimportant by using [property matchers](https://facebook.github.io/jest/blog/2018/05/29/jest-23-blazing-fast-delightful-testing.html#snapshot-property-matchers)

RGBboy: Interesting. I will take a look. Do you think it is worthwhile looking now or as something after this refactor?

markholland: It shouldn't block this but definitely check it out.

PR #392, opened 2018-06-07
Bump App version to v2.0.0

markholland:
We will merge this last when we know we are doing a live release.


RGBboy:
I was wondering why the merge button is green without a review, then I realised this is not a PR against master. Do you think we should require reviews against this branch?

markholland:
@RGBboy Maybe we should just merge the release-pipeline branch and have everything point at master. There's no downside just that any fixes to the release steps would be another PR. We have verified the iOS side at this point though so any fixes would only be on Android.

PR #398, opened 2018-06-07
Refactor event filter selectors

RGBboy:
Related to #345 and the wider refactor to enable memoized reselectors. This PR starts to flatten our selectors so that we can memoize at each level instead of at the state level. This particular PR should give us enough of a performance boost that we can consider releasing.

Further PRs will be made to move our components to the new Event data model and remove the deprecated code;


declanslevin:
Didn’t encounter any regressions in behaviour for anything. 
Looks great on iOS, everything is really snappy and responsive. However I saw some things were still a little sluggish on Android, specifically around Filtering and Favouriting events. I don't think any of them are showstoppers so I will investigate some more and log these as issues if needs be.

PR #399, opened 2018-06-08
Default/placeholder location card

nicollecastrog:
## Test plan

Open "Parade" screen and see that default location card is in the correct position as per screenshot above ^

## Tester check-list

  * [ ] Galaxy S7 (Android 6.0)
  * [ ] iPhone X (iOS 11.x)
  * [ ] iPhone 8 (11.x)
  * [ ] iPhone 7 (iOS 10.x)

  Optional:

  * [ ] Google Pixel 2
  * [ ] Galaxy S8 Plus
  * [ ] Galaxy S7 Edge
  * [ ] Galaxy S6
  * [ ] iPhone 7+
  * [ ] iPhone 6

* [ ] Large Text (=<200%)

* [ ] Airplane mode
* [ ] 2G/3G Network Settings


src/screens/ParadeMapScreen/LocationCard.js
```diff
@@ -0,0 +1,92 @@
...
+import { Image, StyleSheet, View } from "react-native";
+import ContentPadding from "../../components/ContentPadding";
+import Text, { scaleWithFont } from "../../components/Text";
+import locationCardDefault from "../../../assets/images/locationCardDefault.png";
+import { blackTwentyColor, whiteColor } from "../../constants/colors";
+
+type Props = {
+  name?: string,
+  text?: string,
+  image?: any
```
nicollecastrog: it's not my intention to leave this `any` here, I'm just unsure as to how to type it so it support the (eventual) url string from contentful + the local file source referenced in the `defaultProps`...

RGBboy: I would just type this to be the local asset for now. If this is still a problem I recommend using `mixed` instead of `any` (if flow is ok with that). Alternatively, because this is the first time this is being used you could completely omit passing the image and just hard code it in the component for now.

declanslevin:
Found a few accessibility issues but they aren't major so will log these separately, this can be merged

Issue #351, opened 2018-06-01
Saving Events: When saving on Event details page, the vibration and heart animation is repeated when returning to the Event listing page


declanslevin:
## Description of issue

When you open an event from the Event Listing page and tap the heart icon to save it before closing the event, the heart animation and vibration repeats on return to Event Listing page.

Doesn't happen if you open the event from the Featured events on the home page. 

## Expected behaviour

The heart animation and vibration are only triggered the first time you tap on the heart icon. 

## Steps to reproduce

- Open an event from the Event Listing page
- Tap the heart icon to save
> Animation and vibration occur
- Close the Event details page
> Animation and vibration repeat

## OS / Browser / Device model

iPhone 7+
Galaxy S8

RGBboy:
I imagine this is related to the animation code in the SavedEventButton component. Seems like these are being reinstantiated, causing the vibrate to happen.

RGBboy:
If this is too much work, I suggest disabling vibration temporarily.

RGBboy:
I created a PR to disable haptic temporarily: #405.

Looking deeper into this issue, there is a delay in the event list components having their `getDerivedStateFromProps` static method being called causing this to happen. This may be from the changes to support animating unsaved events on the saved events screen or from the animation delay from the navigation library. Surprisingly the number of events in the list affects this. The more events, the slower `getDerivedStateFromProps` takes to get called.

PR #405, opened 2018-06-08
disable haptic temporarily

RGBboy:
As a temporary "fix" for #351 this PR disables haptic feedback on tapping the save event button. Looking deeper into the issue, there is a delay in the event list components having their `getDerivedStateFromProps` static method being called causing this to happen. This may be from the changes to support animating unsaved events on the saved events screen or from the animation delay from the navigation library. Surprisingly the number of events in the list affects this.


RGBboy:
Closing in favor of #409

PR #406, opened 2018-06-08
Minor changes to iOS release process

markholland:
- [x] better slack message
- [x] auto release after passing review
- [x] fix version not being set in Connect

### DIFF
--- a/ios/fastlane/Fastfile
+++ b/ios/fastlane/Fastfile
@@ -125,7 +125,7 @@ platform :ios do
     # Post to Slack
     if defined?(options[:slack_webhook])
       slack(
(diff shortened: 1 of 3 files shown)


ios/fastlane/Fastfile
```diff
@@ -120,12 +120,14 @@ platform :ios do
   lane :promote_to_production do |options|
     latestBuildNumber = latest_testflight_build_number
     # Config in DeliverFile
-    deliver
+    deliver(
```
markholland: If the version isn't explicitly set when calling deliver then a new version won't be created in Connect.

PR #409, opened 2018-06-09
Fix repeat save event animation and vibration

### LINKED ISSUE #351: Saving Events: When saving on Event details page, the vibration and heart animation is repeated when returning to the Event listing page

RGBboy:
Fixes #351. This PR refactors the SaveEventButton to immediately trigger animation and vibration on press and track an animating state. I also ran into an issue with delayed rendering for various screens due to the use of `isFocused` in the respective `shouldComponentUpdate`. To fix this I refactored by creating a `OnlyUpdateWhenFocused` component and wrapped all out screens with this in `src/Navigation.js`.

## Tester check-list

  * [ ] Galaxy S7 (Android 6.0)
  * [ ] iPhone 7 (iOS 10.x)


src/screens/SavedEventListScreen/component.test.js
```diff
@@ -164,29 +160,28 @@ describe("SavedEventListScreen Component", () => {
...
-      const nextProps = {
-        isFocused: false
-      };
-
-      const shouldUpdate = output.instance().shouldComponentUpdate(nextProps);
-
-      expect(shouldUpdate).toBe(false);
-    });
-  });
+  // describe("#shouldComponentUpdate", () => {
```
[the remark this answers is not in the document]
RGBboy: yep

src/screens/SavedEventListScreen/component.js
```diff
@@ -23,11 +23,10 @@ export type Props = {
...
 };
 
-type AllProps = Props & NavigationProps;
-
-class SavedEventListScreen extends Component<AllProps> {
-  shouldComponentUpdate(nextProps: AllProps) {
-    return nextProps.isFocused;
+class SavedEventListScreen extends Component<Props> {
+  shouldComponentUpdate(nextProps: Props) {
+    // may want to omit navigation from this check
```
frigus02: It could be important to do the navigation check here. Otherwise, whenever you change save/unsave an event it will animate here.

src/components/SaveEventButton.js
```diff
@@ -57,16 +49,29 @@ export default class SaveEventButton extends React.Component<Props, State> {
...
+        ReactNativeHapticFeedback.trigger("impactHeavy");
+      }
+
+      this.setState({ animating: true });
     }
-  }
+  };
+
+  completeAnimation = () => {
+    this.setState({ animating: false });
```
frigus02: That could crash when the component was unmounted while the animation is running.

RGBboy: Yeah, I added a test for this and fixed it.

src/screens/SavedEventListScreen/component.js
RGBboy: I think I will refactor in another PR to pull out the calls to navigate and goBack so that we can get referential equality. That would mean all pure components would work correctly.

RGBboy:
If someone could test out the vibration, that would be amazing. I don't have a device which does this.

robbiemccorkell:
Works well for me on iPhone 7

frigus02:
Tested on Pixel 2. Fix works :+1:

PR #427, opened 2018-06-12
Calendar styling fixes

### LINKED ISSUE #419: Update calendar view styling
Just to capture outstanding work to be done for #378, Calendar slide in view:

> Can we please update with following on the calendar screen:

> H2 styling, #lightNavyBlue, centred for the months
> knock back the past inactive dates to #mediumGrey

> Also, after selecting the first date it would be useful to not show the hyphen (but for it to appear if the user selects a second date and to show the range - as it currently is)

abinocoe:
simulator screen shot - iphone 7 - 2018-06-12 at 15 26 43

Fixes #419 

Test adjustments included deletion of a couple of build tests due to this shortcoming in enzyme: https://github.com/facebook/react/issues/8305
the cases in question are covered in terms of the elements being confirmed as disabled in other tests in the suite.

## Tester check-list

  * [ ] Galaxy S7 (Android 6.0)
  * [ ] iPhone X (iOS 11.x)
  * [ ] iPhone 8 (11.x)
  * [ ] iPhone 7 (iOS 10.x)

  Optional:

  * [ ] Google Pixel 2
  * [ ] Galaxy S8 Plus
  * [ ] Galaxy S7 Edge
  * [ ] Galaxy S6
  * [ ] iPhone 7+
  * [ ] iPhone 6


src/screens/DateFilterScreen/DateRangePickerDay.js
```diff
@@ -38,17 +38,17 @@ type DayMarking = {
 };
 
 type DayProps = {
-  state?: "disabled",
+  disabled: boolean,
```
RGBboy: I think you can get rid of `disabled` here now that you calculate it based on todays date. You can then also remove the defaultProps.

src/screens/DateFilterScreen/DateRangePickerDay.test.js
```diff
@@ -72,8 +72,7 @@ describe("dateRangePickerDay", () => {
   it("renders a disabled day", () => {
```
RGBboy: I think this test is now redundant and can be removed.

src/screens/DateFilterScreen/DateRangePickerDay.test.js
```diff
@@ -86,14 +85,17 @@ describe("dateRangePickerDay", () => {
...
       date: date(2018, 7, 12),
-      marking: {},
-      state: "disabled"
+      marking: {}
     });
 
     expect(output.props().disabled).toEqual(true);
     expect(output.props().accessibilityTraits).toContain("disabled");
     expect(output.children().props().style).toEqual(
-      expect.arrayContaining([expect.objectContaining({ opacity: 0.5 })])
```
RGBboy: You could just replace these with a snapshot test.

src/screens/DateFilterScreen/DateRangePickerDay.test.js
```diff
@@ -86,14 +85,17 @@ describe("dateRangePickerDay", () => {
 
     const output = render({
       date: date(2018, 7, 12),
-      marking: {},
-      state: "disabled"
+      marking: {}
     });
```
RGBboy: I recommend mocking the now function in this test so that it is clear that rendering a date before the one returned by now results in a disabled Day component

RGBboy:
Let me know if any of my comments are unclear. I am happy to come chat in person about them.

src/screens/DateFilterScreen/DateRangePickerDay.test.js
abinocoe: It is mocked, I found an implementation in the same file so thought it best to keep the style consistent. The mock is within the test body :)

RGBboy: Apologies, I just saw this was mocked :)

RGBboy:
@abinocoe we made a decision to update with the requested changes so we can cut a release tonight. I have made the update 👍

PR #432, opened 2018-06-12
Rendering optimisation

RGBboy:
So I found an issue with how we thought we were stopping rendering from happening across our screens. We created a helper component that would listen to navigation events and track if it should enable updates or not. Unfortunately this was put outside of the connect, which would still trigger an update when our redux state changes. This PR moves the wrapper to the inside of the connect and now successfully inhibits renders on screens that are not currently focused.

Before merging, it would be great to check on Android as this should make a significant difference to render performance, in particular when filtering events (e.g. selecting dates and categories). Rendering the event list is still pretty heavy (~250ms on iOS in the test environment when looking at the JS performance metrics). It looks like the main issue is to do with React reconciliation.


RGBboy:
I noticed that on the Saved Events screen, after saving an event, the screen animates from the bottom. Not sure exactly why, but this should probably be address before this goes in.

robbiemccorkell:
Tried it on my phone and it seemed like a massive improvement on navigation, filtering and saving of events.

The saved events screen appearing from the bottom briefly I’ve seen on master. I don’t think it’s a consequence of this change

markholland:
Animation from bottom in saved screen is consistently reproducible. Haven’t been able to reproduce in master

RGBboy:
This is likely masked in master because we trigger a re-render of the SavedEvents screen right away. This PR delays it to when it comes into focus. I believe it is caused when saved events transitions from an empty screen to one with events. I'll have a look and see if there is anything we can do in the List component to stop animating in this case.

robbiemccorkell:
Strange, I just tried and I can recreate it from the production build on my phone at the moment.

frigus02:
I think this is a slight performance improvement on Android (Pixel 2) when selecting dates. But it's really hard to tell for me. It's definitely not worse, though. So this should be good.

RGBboy:
I found the issue with the SavedEventsScreen. It is caused by react-navigation and the `createBottomTabNavigator` no longer supporting the `animationEnabled` option. The reason we were not experiencing this before was because we would re-render the saved events screen when any state it relied on changed, even though it was not visible to the user. Unfortunately there is not a simple fix as the `createBottomTabNavigator` no longer has a way to control the animation.

For now I have removed wrapping the SavedEventsScreen with `onlyUpdateWhenFocused`. This has a small amount of performance impact as the screen is now re-rendered, however we gain a lot from applying this to the other screens.

src/components/OnlyUpdateWhenFocused.js
```diff
@@ -1,6 +1,6 @@
 // @flow
-import { Component as ReactComponent } from "react";
-import type { Node } from "react";
+import React, { Component as ReactComponent } from "react";
```
markholland: Is this change needed?

RGBboy:
I made another PR that is slightly different than this one, but makes a bit more of a performance boost for the filter screens: #451. I suggest we consider that change above this one.

Issue #340, opened 2018-05-29
Filtering: Add a "Reset all filters" button to the Events page


declanslevin:
## Description of issue

Currently if you have filter(s) applied it requires a minimum of 2 taps to reset to the default state. If you have multiple filters enabled then it's even more, requiring you to open each filter individually to reset them to the default state. 

## Expected behaviour

When filter(s) are set a "Reset all filters" button appears on the Events page.

clmntnbrn:
![event listing reset filters](https://user-images.githubusercontent.com/12392657/40920352-2dde0e94-6804-11e8-8289-4825fcc8e795.jpg)
Display Reset all filters once one has been set - font size to be the same as 'Select dates' etc

jonyardley:
I will pick this up today :)

robbiemccorkell:
If the header is growing, does that mean the list area will be super tiny on an iPhone 4?

robbiemccorkell:
If so we might be able to mitigate that in the future by hiding the header like in #66

clmntnbrn:
yep @robbiemccorkell I reckon that's a good shout - only the date remains sticky. Let's add the reset button and I'll talk to @stevecottle about whether we should do the rest

charypar:
I appreciate I'm a bit too late to the party, but would this look better as a thin sticky banner saying something like "Filters applied" at the top of the list with a ✖️icon on the right? Like one of the date headers, but a different colour, maybe the green we use for the number badge. 

This look like a bit of an afterthought and will make a lot of the UI move in an odd way when tapped.

PR #435, opened 2018-06-13
Reset all filters button on Events page

jonyardley:
This PR adds a `Reset all filters` button to the `Events` page.

The button should exhibit the following behaviours:
* When no filters are selected it should not be visible
* When one or more filters are selected it should be visible
* When pressed it should reset all filters and fade out and up
* When pressed it should also reset the Event List view position to the top

## Test Plan
Following the steps outlined above should suffice.

## Areas of risk
* Events page
  * All Filters
  * Event List

## Further Notes
We have agreed that the header should probably be hidden when scrolling however due to a funky jitterbug with the scrolling and the refresh component we have decided to pull this feature out into a new issue/PR so that it can be tackled separately in order to get this out sooner.

Fixes: https://github.com/redbadger/pride-london-app/issues/340

## Pre-flight check-list

Before raising a pull request

* ~~Documentation~~
* [x] Unit tests

## Pre-merge check-list

* [x] Test plan provided
* [ ] Tester approved

## Tester check-list

  * [x] Galaxy S7 (Android 6.0)
  * [x] iPhone X (iOS 11.x)
  * [x] iPhone 8 (11.x)
  * [ ] iPhone 7 (iOS 10.x)

  Optional:

  * [ ] Google Pixel 2
  * [ ] Galaxy S8 Plus
  * [ ] Galaxy S7 Edge
  * [ ] Galaxy S6
  * [x] iPhone 7+
  * [ ] iPhone 6

* [ ] Large Text (=<200%)

* [ ] Airplane mode
* [ ] 2G/3G Network Settings


src/actions/event-filters.js
```diff
@@ -61,4 +61,4 @@ export const clearStagedEventFilters = () => (
 
 export const clearEventFilters = () => (
   dispatch: Dispatch<EventFiltersAction>
-) => dispatch({ type: "CLEAR__EVENT_FILTERS" });
```
jonyardley: 🎺

src/screens/EventsScreen/FilterHeader.js
```diff
@@ -20,31 +20,54 @@ export type Props = {
...
+                  onPress={resetAllFiltersPress}
+                />
+              </View>
+            )}
+            <View testID="event-filter-header" style={styles.content}>
+              <FilterHeaderCategories
+                onFilterPress={onFilterCategoriesPress}
+                selectedCategories={selectedCategories}
+              />
+            </View>
```
jonyardley: I think I might pull this out into a separate component. Enough logic in here to make this a separate thing.

src/screens/EventsScreen/ResetAllFiltersButton.js
```diff
@@ -0,0 +1,113 @@
...
+    this.fadeOut();
+  };
+
+  resetAnimation = (): void => {
+    this.topOffset = new Animated.Value(DEFAULT_TOP_OFFSET_VALUE);
+    this.fadeValue = new Animated.Value(DEFAULT_FADE_VALUE);
+  };
+
+  animationFinished = (): void => {
+    this.setState({ isAnimating: false });
```
frigus02: This might be called when the button has already been unmounted, in which case `setState` will throw an exception.

src/screens/EventsScreen/ResetAllFiltersButton.js
```diff
@@ -0,0 +1,113 @@
...
+
+type State = {
+  isAnimating: boolean
+};
+
+const DEFAULT_HEIGHT = 120;
+const DEFAULT_FADE_VALUE = 1;
+const DEFAULT_TOP_OFFSET_VALUE = 0;
+
+class ResetAllFiltersButton extends React.Component<Props, State> {
```
frigus02: Do you want to make this a `PureComponent`, so it doesn't rerender all the time?

src/screens/EventsScreen/FilterHeader.js
```diff
@@ -13,38 +13,59 @@ import {
...
       <View accessibilityTraits={["header"]} style={styles.container}>
         <ContentPadding>
-          <View testID="event-filter-header" style={styles.content}>
-            <FilterHeaderCategories
-              onFilterPress={onFilterCategoriesPress}
-              selectedCategories={selectedCategories}
+          <View>
+            <ResetAllFiltersButton
+              visible={anyAppliedFilters}
+              onPress={() => {
```
frigus02: If you make `ResetAllFiltersButton` a pure component, this shouldn't be an inline function.

frigus02:
The reset button has a weird touch area. It is mostly above the text. Do you think we are able to increase the touch area below the text?

The highlighted part in the screenshot is the touchable area:

image

src/screens/EventsScreen/FilterHeader.js
jonyardley: Could you elaborate a bit? Not sure I follow this point.

src/screens/EventsScreen/ResetAllFiltersButton.js
jonyardley: Certainly!

src/screens/EventsScreen/ResetAllFiltersButton.js
jonyardley: Ah yeah good catch.  I notice that isMounted() is now depricated. Do you have a pattern in mind to catch this? Maybe a try catch (might be a bit crude).

jonyardley:
@frigus02 The padding this will be a bit tricky here. As we will need to animate the padding on the categories header too and would rather keep animation to a minimum.

src/screens/EventsScreen/ResetAllFiltersButton.js
jonyardley: Looked into this a bit more and there is a post from Dan Abramov on this issue:

https://github.com/facebook/react/issues/2787#issuecomment-304485634

```
In any case, React doesn’t throw errors when this happens anymore. But we will intentionally keep the suboptimal case clunky so that the developers have incentives to implement the right solution (cancellation).
```

It is now only a warning and not an error so inclined to leave this as is due do it being a edgy edge case unless you think it will be a major issue?

frigus02: Hm, interesting. We tested this we @RGBboy for the heart animation. I'm pretty sure we could make this crash. 🤔 

As far as I can see the recommendation is to set a field, which is basically what we ended up doing here:

https://github.com/redbadger/pride-london-app/blob/3a94242f84d188f81c9f3f1db238679069d094f0/src/components/SaveEventButton.js#L52-L54

Maybe our crash was unrelated though and we can just leave this at it is.

src/screens/EventsScreen/FilterHeader.js
frigus02: As fas as I know, this will generate a new function every time in `render`, which means that inside `ResetAllFiltersButton` this check will always be false: `props.onPress === nextProps.onPress`. So it will render every time.

We could solve this by adding the function to the `FilterHeader` class:

```js
resetAllFilters = () => {
  this.props.resetAllFiltersPress();
  this.props.scrollEventListToTop();
};
```

Then use `onPress={this.resetAllFilters}`. This will always pass in the same function.

frigus02:
Ah, right. Makese sense with the animation :+1:

src/screens/EventsScreen/ResetAllFiltersButton.js
frigus02: I created a unit test for `unmount` -> `setState` and it threw an exception. So I just added a quick fix for this, just to be on the safe side.

PR #438, opened 2018-06-13
Fix Events Listing: Filtering can break the events list

### LINKED ISSUE #436: Events Listing: Filtering can break the events list
## Description of issue

Rapidly scrolling the events listing with a lot of events listed can break it - events stop loading correctly and date headers get duplicated. The page then stutters quite erratically as it tries to update and load the missing events.

It seems to be triggered when filtering the list with the date and event filters. 

## Expected behaviour

Events are loaded correctly and the page scrolls smoothly when rapidly scrolling. 

## Steps to reproduce

- Open the Events page
- Filter by date, e.g. 13 jun - 20 jul
- Filter by area, e.g. Central, North, South, West
- Scroll down rapidly then back to the top
> Sunday 1 July is duplicated repeatedly

If you clear the filters and scroll the list rapidly it can end up looking like this:
https://drive.google.com/file/d/1D4SmZkGkLVFcIBAbPfRozXoZHElB9EAi/view?usp=sharing

I've reproduced on:
Galaxy S8

Joe was able to reproduce on his iPhone (not sure what model)

RGBboy:
Fixes #436.

So it turns out we were not respecting timezones when creating the ids for the event list groups. This along with some of our test data having obscure timezones would cause repeated keys being used. This PR updates user facing times to always be coerced to the local timezone (while still preserving the actual time, i.e. time from epoch). While this means users would see different times depending on their timezone settings, they would see the correct time relevant to their device clock. Happy to be challenged on this.

## Pre-merge check-list

* [ ] Tester approved

## Tester check-list

  * [ ] Galaxy S7 (Android 6.0)
  * [ ] iPhone X (iOS 11.x)
  * [ ] iPhone 8 (11.x)
  * [ ] iPhone 7 (iOS 10.x)

### DIFF
--- a/src/lib/date.js
+++ b/src/lib/date.js
@@ -24,6 +24,13 @@ const parse = (date: string) => LuxonDateTime.fromISO(date, { setZone: true });
 export const toFormat = (date: string, format: string) =>
   parse(date).toFormat(format);
 
+export const toLocalFormat = (date: string, format: string) =>
+  // Note toLocal transforms the timezone and time but retains
+  // the UTC time. Essentially it keeps the same value
+  parse(date)
+    .toLocal()
+    .toFormat(format);
+
 export const isBefore = (d1: string, d2: string) => +parse(d1) < +parse(d2);
 
 export const now = () =>
--- a/src/components/EventList.js
+++ b/src/components/EventList.js
@@ -9,7 +9,7 @@ import SectionHeader from "./SectionHeader";
 import { whiteColor } from "../constants/colors";
 import type { SavedEvents, Event, EventDays } from "../data/event";
 import {
-  toFormat as formatDate,
+  toLocalFormat as formatDate,
   FORMAT_WEEKDAY_DAY_MONTH,
   FORMAT_YEAR_MONTH_DAY
 } from "../lib/date";
--- a/src/components/EventTile.js
+++ b/src/components/EventTile.js
@@ -5,7 +5,7 @@ import ConnectedImageBackground from "../components/ImageBackground";
 import Text from "../components/Text";
 import CategoryPill from "./CategoryPill";
 import { imageBgColor, eventTileTextColor } from "../constants/colors";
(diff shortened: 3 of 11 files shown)


robbiemccorkell:
So do you mean that time time will display relevant to their current time? Or do you mean it's just the formatting that will be different but we still show London time.

The reason I ask is that we want to show events as the time the event will be on that day in London, not what it would be for the user's locale. Otherwise when planning they would see an event time corrected for their country, and be surprised when the time changes when they arrive in London.

RGBboy:
With this change the time would be corrected to that of the users locale. We could instead force +01:00, in which case I would refactor the `toLocalFormat` to `toLondonFormat` and force `+01:00`.

RGBboy:
❤️ ⏲

RGBboy:
I have refactored to create a `toLondonFormat` function instead to make it clear the expectation;

src/lib/date.js
```diff
@@ -24,6 +24,13 @@ const parse = (date: string) => LuxonDateTime.fromISO(date, { setZone: true });
 export const toFormat = (date: string, format: string) =>
   parse(date).toFormat(format);
 
+export const toLondonFormat = (date: string, format: string) =>
+  // Note setZone transforms the timezone and time but retains
+  // the epoch time. Essentially it keeps the same value
+  parse(date)
+    .setZone("utc+1")
```
frigus02: I think this should be `setZone("Europe/London")`, so it continues to work as expected in the Winter.

frigus02: Okay, we need to check if this is supported on Android and iOS. Appearently IANA time zone names are implemented using built-in JavaScript Intl APIs:

https://moment.github.io/luxon/docs/manual/zones.html#iana-support

frigus02: Of course it doesn't work on Android 😞. So, do we always want to convert to british summer time?

PR #439, opened 2018-06-13
Latest React Native RC

markholland:
Just for reference and testing.


markholland:
164 flow errors 😱

RGBboy:
Most of those errors seem to be originating from node_modules `Cannot resolve module...` not our code...

PR #451, opened 2018-06-14
Rendering optimisation take 2

RGBboy:
This PR takes a slightly different approach than #432 where we instead compute and pass an isFocused prop to our connected components. Based on this, we either pass on a cached return value of mapStateToProps or compute a new one. This should result in the filter screens being much more performant on both platforms.

If this PR is chosen, we will close the #432.


src/components/WithIsFocused.js
```diff
@@ -0,0 +1,55 @@
...
+  navigation: NavigationScreenProp<NavigationState>
+};
+
+type State = {
+  isFocused: boolean
+};
+
+// This component stops children from updating when the given navigation
+// is not inFocus (I.e. not in the state between willFocus and blurred)
+const withIsFocused = <A>(Component: ComponentType<A>) =>
```
RGBboy: I still need to add tests for this HOC. Just want to gauge how effective this is on real Android devices before I do.

src/screens/EventDetailsScreen/index.js
```diff
@@ -61,25 +60,33 @@ const reducePerformancesHelp = (performances: Performances) => (
...
-    isSaved: state.savedEvents.has(id),
-    performances
-  };
+const mapStateToProps = (
+  state: State,
+  { navigation, isFocused }: OwnProps
+): StateProps => {
+  if (!cache || isFocused) {
+    const id = navigation.state.params.eventId;
+    const event = getEventById(state, navigation.state.params.eventId);
```
frigus02: This can be `const event = getEventById(state, id);`.

src/screens/SavedEventListScreen/component.js
```diff
@@ -24,8 +24,16 @@ export type Props = {
...
-    return nextProps !== this.props;
+    // Intentionally do not check this.props.navigation
+    return (
+      nextProps.events !== this.props.events ||
+      nextProps.savedEvents !== this.props.savedEvents ||
+      nextProps.addSavedEvent !== this.props.addSavedEvent ||
+      nextProps.removeSavedEvent !== this.props.removeSavedEvent ||
+      nextProps.loading !== this.props.loading ||
+      nextProps.refreshing !== this.props.refreshing ||
+      nextProps.updateData !== this.props.updateData
```
frigus02: Do we still need this now, that we check `isFocused` in the connector?

frigus02:
This PR brings a massive performance improvement on Android. We should definitelty get this in and close #432.

declanslevin:
Tested on:

Galaxy S8
Galaxy Note Edge
iPhone 7+

On the S8 the performance improvement is immediately apparent - there is still a slight delay, but selecting filters is so much quicker than it was. Saving events is pretty much instant now. 

On the Note Edge, which is a much older device, there are still very noticable delays, but they are a lot shorter than they were previously. 

The iPhone 7+ shows a subjective performance increase as well, though it is hard to quantify as the change is not as dramatic as on Android, due to performance being better to begin with on iPhone.

PR #466, opened 2018-06-18
Change VoiceOver checkbox behaviour

### LINKED ISSUE #322: Accessibility iOS : When a button is toggled with Voiceover enabled, it reads out both the previous state and the new state labels
## Description of issue

When double-tapping on a button to toggle its state with Voiceover enabled on iOS, it reads out both the previous and new state of the button, which is quite confusing as it isn't clear which state the button is now in.

## Expected behaviour

When double-tapping to toggle a button, only the new state label of the button is read out. 

## Steps to reproduce

- Enable VoiceOver
- Open the Category filter
- Tap on "Cabaret and Variety"
> It reads out "Cabaret and Variety, Checkbox, Empty"
- Double-tap on "Cabaret and Variety"
> It reads out "Cabaret and Variety, Checkbox, Empty, Cabaret and Variety, Checkbox, Selected"

It should probably just read out "Selected, Cabaret and Variety" when the label is double-tapped to align with the iOS checkbox accessibility behaviour. 

iPhone 7+

abinocoe:
Current behavior on iOS is reading out the label and status of checkbox(type elements) twice upon toggle when VoiceOver is activated. This pull attempts to bring behavior in line with what is typical on iOS, i.e. voice over of "Selected, <item label>" or just "<item label>" if selected/not selected respectively

On toggle of item (double tap once item has already been selected) read out the new item status only

Fixes #322 

**This will need to be tested on a device as the Xcode simulator does not support VoiceOver.**

* [ ] Text-to-Voice (iOS: Speak Selection)
* [ ] Text-to-Voice (Android: Speak Selection)


frigus02:
It's not quite correct. It still says the label twice when double tapping on an item.

frigus02:
I tested the whole day and this is what I came up with:

* On double tap VoiceOver reads out the current element, even in native apps like the calendar. It doesn't matter if the current element is a button or just text.
* Because in React Native rendering (and therefore changing the accessibility traits) is asyncronous, VoiceOver will first read the old state of the checkbox. Then, because it changes, it reads the new state.

I tried different things like using [`setNativeProps`](http://facebook.github.io/react-native/docs/direct-manipulation.html#setnativeprops-with-touchableopacity) to change the accessibility traits earlier but without any effect.

The only thing that does work is responding to the tap synchronously in native code. That is essentially what is happening in the native `Switch` component. Which is why it works great with VoiceOver.

Unfortunately there is no native check box component. I'm honestly thinking about making one.

If anyone has any other ideas, let me know.

src/components/CheckBox.js
```diff
@@ -16,22 +15,40 @@ type Props = {
...
-
-CheckBox.defaultProps = {
-  style: {}
-};
+class CheckBox extends React.Component<Props> {
+  static defaultProps = {
+    style: {}
+  };
+
+  shouldComponentUpdate(nextProps: Props) {
```
frigus02: We currently use CheckBox only in the filter modal and we use an inline arrow function for `onChange`. This means all check boxes in the list are always rerendered. The easiest fix is to use a custom `shouldComponentUpdate` here.

frigus02:
That's ready to be tested now. It fixes the check boxes in both the category and "price, area, ..." filter screens.

On iOS it should exactly match native check boxes now. I found that some apps on iOS announce check boxes as buttons and some don't. I opted for buttons to indicate that they are clickable.

On Android the check boxes are announced as radio buttons. This is unfortunate but it is the only thing that React Native provides. We do get a good feedback when checking/unchecking them, though. So I think this is fine.

Feedback welcome 🙂

PR #477, opened 2018-06-22
Trello 148 stages on map

abinocoe:
In order to avoid horrific conflicts, this probably shouldn't go much further without the user location branch being merged, due to the separation of concerns in the map component

https://trello.com/c/v6y8b2CC/148-parade-day-show-stages-on-map

## Tester check-list

  * [ ] Galaxy S7 (Android 6.0)
  * [ ] iPhone X (iOS 11.x)
  * [ ] iPhone 8 (11.x)
  * [ ] iPhone 7 (iOS 10.x)

  Optional:

  * [ ] Google Pixel 2
  * [ ] Galaxy S8 Plus
  * [ ] Galaxy S7 Edge
  * [ ] Galaxy S6
  * [ ] iPhone 7+
  * [ ] iPhone 6

* [ ] Large Text (=<200%)

* [ ] Airplane mode
* [ ] 2G/3G Network Settings


RGBboy:
I'm conscious that changes in #408 will change the Map part of this PR. I'm going to hold off reviewing this until that is in 👍

src/screens/ParadeMapScreen/Map.js
```diff
@@ -0,0 +1,192 @@
...
+            strokeWidth={5}
+            strokeColor={warmPinkColor}
+            lineJoin="bevel"
+          />
+          <Marker
+            coordinate={{ longitude: -0.14223, latitude: 51.51616 }}
+            centerOffset={{ x: 0, y: -15 }}
+            stopPropagation
+          >
+            <View style={styles.markerView}>
```
markholland: Should this be its own component? look to be repeated a few times here

abinocoe: This repetition is reduced in the user location PR

src/screens/ParadeMapScreen/Map.js
```diff
@@ -37,12 +41,19 @@ const shouldNeverAsk = (status: PermissionStatus) => status === "restricted";
 type Props = {
   route: Array<Coordinates>,
   paradeRegion: Region,
-  terminals: Array<Terminals>
+  terminals: Array<Terminals>,
+  stages: Array<Event>,
+  savedEvents: SavedEvents,
+  addSavedEvent: string => void,
+  removeSavedEvent: string => void,
+  onPress: (id: string) => void
```
frigus02: A better name for this might be `onStagePress`. A simple `onPress` on a `Map` component suggests that is called when the map itself is pressed.

src/screens/ParadeMapScreen/Map.js
```diff
@@ -110,6 +123,14 @@ class Map extends Component<Props, State> {
     }
   };
 
+  handleMarkerPress(stage: Event) {
+    this.setState({ tileDetails: stage, activeMarker: stage.id });
+  }
+
+  handleMapPress = () => {
+    this.setState({ tileDetails: null, activeMarker: null });
```
frigus02: I think we should check here if `tileDetails` and `activeMarker` are already null and don't call `setState` in this case. [Reason](https://reactjs.org/docs/react-component.html#setstate):

> `setState()` will always lead to a re-render unless `shouldComponentUpdate()` returns `false`.

src/screens/ParadeMapScreen/Map.test.js
```diff
@@ -1,18 +1,31 @@
...
+      {...props}
+      stages={[stages]}
+      addSavedEvent={() => {}}
+      removeSavedEvent={() => {}}
+      onPress={() => {}}
+      savedEvents={new Set()}
+    />
+  );
+
+const stages = sampleOne(generateEvent, { seed: 5728 });
```
frigus02: This is one `stage`, isn't it? 🙂

src/screens/ParadeMapScreen/Map.js
frigus02: Hm, or maybe it's fine. That's really hard to understand from the documentation. I think we would at least have to make `Map` a `PureComponent`.

src/screens/ParadeMapScreen/Map.test.js
abinocoe: lol, yes. I will change

frigus02: It caught me a few lines above where it is used as `stages={[stages]}`. I thought: why do we an array in an array? 😄

src/screens/ParadeMapScreen/Map.js
abinocoe: I'm confused, you would like me to make Map a PureComponent?

src/screens/ParadeMapScreen/Map.test.js
frigus02: Hm, you did change it. Just not in this file. It's fine, though. 😄

abinocoe: Oh dear me! I have changed it here now too 😅

PR #479, opened 2018-06-23
Bundle more recent version of Javascript Core(jsc) on Android

markholland:
When using React Native, on iOS the platform jsc is used and on Android react native bundles it's own version.

This means that every new version of iOS will provide a "better" version of jsc to React Native apps while on Android the version being bundled is from 2014 and unlikely to change anytime soon. [link](https://github.com/facebook/react-native/issues/10245)

The community has responded by providing pre-built versions of jsc that are more up to date by distributing them through npm. That is what this PR uses.

One downside is that the minSdk has to be raised to 21 as this is a requirement of the npm package used.

Opening this for testing and feedback as to whether it provides any noticeable improvements.

## Pre-merge check-list

* [ ] Test plan provided
* [ ] Tester approved

## Tester check-list

  * [ ] Galaxy S7 (Android 6.0)
  * [ ] iPhone X (iOS 11.x)
  * [ ] iPhone 8 (11.x)
  * [ ] iPhone 7 (iOS 10.x)

  Optional:

  * [ ] Google Pixel 2
  * [ ] Galaxy S8 Plus
  * [ ] Galaxy S7 Edge
  * [ ] Galaxy S6
  * [ ] iPhone 7+
  * [ ] iPhone 6

* [ ] Large Text (=<200%)

* [ ] Airplane mode
* [ ] 2G/3G Network Settings

### DIFF
--- a/package.json
+++ b/package.json
@@ -28,6 +28,7 @@
     "bugsnag-react-native": "^2.9.3",
     "contentful": "^5.0.5",
     "core-js": "^2.5.6",
+    "jsc-android": "^224109.0.0",
     "lottie-react-native": "2.3.2",
     "luxon": "^1.2.0",
     "ramda": "^0.25.0",
--- a/android/app/build.gradle
+++ b/android/app/build.gradle
@@ -222,6 +222,12 @@ repositories {
     maven { url 'https://maven.fabric.io/public' }
 }
 
+configurations.all {
+    resolutionStrategy {
+        force 'org.webkit:android-jsc:r224109'
+    }
+}
+
 dependencies {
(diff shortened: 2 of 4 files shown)


package.json
```diff
@@ -28,6 +28,7 @@
     "bugsnag-react-native": "^2.9.3",
     "contentful": "^5.0.5",
     "core-js": "^2.5.6",
```
RGBboy: Does this mean pulling in core-js can go? We added this in order to support some of the new String prototype methods.

markholland: I've been trying to check this all week but could never get Android(on any branch) to run on my laptop. Trying on desktop now.

PR #499, opened 2018-06-29
Update dev dependencies

frigus02:
A bit of housekeeping. Update dev dependencies to latest versions.

* ESLint needs to wait because the airbnb config is not updated, yet.
* Flow and React are bound to the React Native version.
* Detox will get it's own PR.


package.json
```diff
@@ -54,17 +49,17 @@
...
+    "babel-jest": "23.2.0",
     "babel-preset-react-native": "4.0.0",
     "bugsnag-sourcemaps": "^1.0.3",
-    "commander": "^2.13.0",
-    "contentful-management": "^4.1.4",
-    "detox": "^7.3.3",
-    "dotenv": "^5.0.0",
+    "commander": "^2.16.0",
+    "contentful-management": "^5.1.4",
+    "detox": "^8.0.0",
```
markholland: I was planning on updating detox in its own PR as there some cool new stuff to reduce boilerplate in our tests

frigus02: Ah nice. I’ll remove detox from this one then.

PR #521, opened 2018-06-29
Don't set a valid bugsnag api key when running e2e

markholland:
Still investigating as to why this only seems to have happened in master and not during the PR branch.

No need for e2e job to ever talk to Bugsnag so we set an invalid api key.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -462,9 +462,9 @@ jobs:
     <<: *job-defaults-macos
     working_directory: /Users/distiller/project
     environment:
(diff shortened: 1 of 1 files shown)


markholland:
Well this clearly doesn't work as the stage being sent by e2e is "production"
https://github.com/redbadger/pride-london-app/blob/master/src/index.js#L25

markholland:
This doesn't work and I've run out of ideas as to why

markholland:
Closing until working

Issue #502, opened 2018-06-29
Android (6.0.1) You can spam the location button on the map.


AndreiDanielIonescu:
## Description of issue
You can spam the location button on the map.
Describe the error in as much detail as possible.
If you receive the message for not being able to pick up the location, you can actually spam the button a lot so the message will be flickering on your screen
## Expected behaviour

If it's a functional issue, describe what would have been the correct behaviour.
 Is it possible to put a timeout on the button, a delay to wait for a response?
## Steps to reproduce
 Go to map.
 Spam the location button.

List the exact steps you took to arrive at the issue.
 The message for not being able to locate you should have a delay and not flicker 

## OS / Browser / Device model
 Android 6.0.1

List all the devices and browsers effected, with specific version numbers.

## Screenshots

If it's a visual bug, please provide screenshots or screen captures.

frigus02:
Sounds like the same as #501

RGBboy:
I will be looking into this along with #501 and #503.

PR #540, opened 2018-07-01
Refactor current location

### LINKED ISSUE #501: Map: My location button takes me to my location twice when I press it twice
## Description of issue

On Android:

- Go to parade map and click on location button.
- Before it takes you your location, click it again.
- Wait until it takes you to your location and move away.
- It will now take you to your location again.

Possibile cause:

Every click on the "my location" button seems to trigger a call to the location service. On Android it can take a few seconds to receive the location. Once it gets it, it moves the map. In between you can still move the map to anywhere you want, though.

## Expected behaviour

- It should only ask the location service once. When I click the button again and it didn't receive something from the last call, don't call again.
- If the users moves the map while we retrieve the location, we should not go to the users location anymore??

Most visible on Android.

## Screenshots

If it's a visual bug, please provide screenshots or screen captures.

RGBboy:
This PR fixes #501 #502 and #503 along with a couple of other unreported bugs that the current location exhibited.

For the problems we were facing, refactoring to use Observables seemed like the right fit. Essentially we want the ability to stream location and permission updates in order to display the correct interface on the map. For example:

* when location permission is restricted, we should not be displaying the moveToLocation icon.
* when location permission is authorized, we should automatically retrieve it to avoid another round trip when the user triggers moveToLocation.
* when location permission is denied or undetermined, we should prompt the user for their permission then stream location updates if authorized.

As a result I have created `lib/geolocation` which exposes `passiveLocationStream` and `activeLocationStream` to enable us to subscribe to a stream of LocationStatus values. LocationStatus can be one of the following:

```
Authorized + Awaiting
Authorized + Tracking(x, y)
Authorized + Error
Checking
Denied
Requesting
Restricted (only applies to Android users who request to never be asked)
Undetermined
```

`passiveLocationStream` automatically streams location updates if permissions allow this to be done without prompting the user. For example the output may be:

```
|-A---B---C---D---E
```
Where:
```
A = Checking
B = Authorized + Awaiting
C = Authorized + Tracking(1, 1)
D = Authorized + Tracking(1, 2)
E = Authorized + Error
```

`activeLocationStream` will prompt the user to authorize accessing location information. If authorized it will automatically stream location updates. For example the output may be:

```
|-A---B---C---D---E
```
Where:
```
A = Requesting
B = Authorized + Awaiting
C = Authorized + Tracking(1, 1)
D = Authorized + Tracking(1, 2)
E = Authorized + Error
```

The user is prompted between the `A` and `B` values.

## Tester check-list

  * [ ] Galaxy S7 (Android 6.0)
  * [ ] iPhone X (iOS 11.x)
  * [ ] iPhone 8 (11.x)
  * [ ] iPhone 7 (iOS 10.x)

  Optional:

  * [ ] Google Pixel 2
  * [ ] Galaxy S8 Plus
  * [ ] Galaxy S7 Edge
  * [ ] Galaxy S6
  * [ ] iPhone 7+
  * [x] iPhone 6

* [ ] Large Text (=<200%)

* [ ] Airplane mode
* [ ] 2G/3G Network Settings


markholland:
Installing the build after syncing with master fixed the empty settings issue.

markholland:
My comments aren’t related to this PR as I’ve been able to reproduce them in master. This was my first time trying the location and the issues were quite jarring.

RGBboy:
@markholland Would the following behavior do what you expect:

New app (no permission preference):
```
click location button
-> prompt for permission
-> user denies permission
-> location button disappears + Does not navigate to user
```

Existing app with denied permission
```
click location button
-> prompt for permission
-> user deny permission
-> location button disappears + Does not navigate to user
```

If so, when would you ever expect to be taken to the permission screen?

markholland:
@RGBboy Why would the location button disappear?

Ok, I think I'm understanding the decision here. A pop up of our own before taking me to settings would be better though and obviously out of scope of this PR.

The main issue is that denying permission still takes me out of the app.

RGBboy:
This is partly to do with the face that Android has an extra permission state, Restricted, that iOS doesn't. This is the value Android returns when a user chooses never to ask again.

I think we have a few options to make this better:

1. Transform the Denied message to a Restricted one when listening to the activeLocationStream. This would mean when a user denies access, the location button disappears and they will not be able to select it again. However, if the user navigates away from the map and back again, they will be presented the navigate to location button again. If they then click the navigate to location they would then be taken to the settings screen.

2. We remove taking iOS users to the settings screen. Instead, they would just be presented with the same option pop up when pressing the navigate to location button. Even if the user denies access, the button remains pressable.

3. A combination of the above. Remove taking iOS users to the settings screen and also map the denied to restricted for activeLocationStream. This would mean that users that deny the pop up will not be presented the button. However when they navigate away and back to the map it will show again, and present the pop up again if pressed.

Personally I think the last option is actually the better of the three. We only ask if they intended. And no longer take them away from the app.

RGBboy:
Just been looking at this a but deeper and the behaviour on the two platforms has some further difficulties. On iOS, if a user has denied access, they will no longer see a pop up ever. Hence why we were navigating to the settings screen. Android however has three options, Authorized, Declined and Restricted. Authorized grants us access, declined means we need to ask again via pop up, and restricted means we never ask.

RGBboy:
Hopefully I have addressed the major problem on iOS now. I have moved the Linking to settings screen into the geolocation lib. This will now happen if both the passiveLocationStream, and the activeLocationStream return denied (indicates that the user has denied access via pop up). This needs thorough testing on Android because some status remapping may affect that platform.

markholland:
The behavior on iOS is what I would expect but the copy of the dialogue when not allowed could be better and include a indication of what needs to be done for it to work rather than "try again later" which will never help.

RGBboy:
My latest changes address the copy in the dialog problem. We now present the user a choice to go to the settings screen if they have previously denied access. The other dialog will still show if the users location is authorized, but we can not get it for some technical reason and they click the navigate to location marker.

PR #549, opened 2018-07-02
497 ios stage focus

### LINKED ISSUE #492: Stages on Map: Pin doesn't centre to screen when tapped on iOS
## Description of issue

When I tap on a stage pin the pin remains in its current position

## Expected behaviour

When I tap on a stage pin I'd expect the pin to move to the center of the screen / frame

## Steps to reproduce

Navigate to Parade > Map
Drag screen to make a stage pin off-center
Tap on stage pin
Stage pin remains in position

Jun's iPhone X

## Screenshots

none

abinocoe:
Fixes #492 and extends behaviour to include all markers on map, to bring iOS in line with android

Expected behaviour: when a marker on the map is clicked, the map will move to centre on that marker

## Tester check-list

  * [ ] Galaxy S7 (Android 6.0)
  * [ ] iPhone X (iOS 11.x)
  * [ ] iPhone 8 (11.x)
  * [ ] iPhone 7 (iOS 10.x)

  Optional:

  * [ ] Google Pixel 2
  * [ ] Galaxy S8 Plus
  * [ ] Galaxy S7 Edge
  * [ ] Galaxy S6
  * [ ] iPhone 7+
  * [ ] iPhone 6

* [ ] Large Text (=<200%)

* [ ] Airplane mode
* [ ] 2G/3G Network Settings


src/screens/ParadeMapScreen/Map.test.js
```diff
@@ -172,3 +172,40 @@ describe("moveToCurrentLocation", () => {
     });
   });
 });
+
+describe("handleMarkerPress", () => {
+  it("updates state with stage marker details", () => {
```
RGBboy: I think there may be a test earlier in the file that is the same as this. That test could be removed.

src/screens/ParadeMapScreen/__snapshots__/Map.test.js.snap
```diff
@@ -580,10 +581,17 @@ exports[`Map component renders correctly 1`] = `
           },
         ]
       }
+      mapRef={
```
RGBboy: I'm confused as to why mapRef is being passed to the AmenityMarkers props. Are we spreading props?

src/screens/ParadeMapScreen/__snapshots__/Map.test.js.snap
```diff
@@ -1289,10 +1298,17 @@ exports[`Map component renders without location button when permission is restri
...
```
RGBboy: I'm confused as to why mapRef is being passed to the TerminalMarkers props. Are we spreading props?

src/screens/ParadeMapScreen/Map.test.js
RGBboy: Line 105, test "updates tileDetails and activeMarker state when stage is clicked" is the one that should be bale to be removed.

src/screens/ParadeMapScreen/__snapshots__/Map.test.js.snap
abinocoe: This is my bad, I neglected to update the snaps after removing a ref I was passing that I was gonna use but ended up not needing to

src/screens/ParadeMapScreen/__snapshots__/Map.test.js.snap
abinocoe: ditto ^^

PR #560, opened 2018-07-02
Message Bar component

abinocoe:
This displays an animated banner to alert users when current location cannot be found, to replace the current alert dialog.

Banner will drop down and persist for 7 seconds before sliding back up, unless the banner is tapped on, when the slide up will happen immediately.

screen shot 2018-07-02 at 17 07 35

* [ ] Tester approved

## Tester check-list

  * [ ] Galaxy S7 (Android 6.0)
  * [ ] iPhone X (iOS 11.x)
  * [ ] iPhone 8 (11.x)
  * [ ] iPhone 7 (iOS 10.x)

  Optional:

  * [ ] Google Pixel 2
  * [ ] Galaxy S8 Plus
  * [ ] Galaxy S7 Edge
  * [ ] Galaxy S6
  * [ ] iPhone 7+
  * [ ] iPhone 6

* [ ] Large Text (=<200%)

* [ ] Airplane mode
* [ ] 2G/3G Network Settings


src/screens/ParadeMapScreen/Map.js
```diff
@@ -189,6 +187,13 @@ class Map extends PureComponent<Props, State> {
 
     return (
       <View style={styles.mapWrapper}>
+        {this.state.userLocation.type === "authorized" &&
+          this.state.userLocation.location.type === "tracking" && (
```
RGBboy: I believe this should be `error` rather than `tracking`. Currently this message is displaying if we can find the users location.

src/screens/ParadeMapScreen/Map.js
```diff
@@ -173,6 +169,8 @@ class Map extends PureComponent<Props, State> {
 
   // $FlowFixMe
   mapViewRef: ElementRef<typeof MapView> = React.createRef();
+  // $FlowFixMe
+  messageBannerRef: ElementRef<typeof MessageBanner> = React.createRef();
```
RGBboy: I think this can be removed.

src/screens/ParadeMapScreen/Map.js
RGBboy: One minor bug that I can see happening, is if the state of userLocation changes out of the authorized error state before the MessageBanner animates out, it will disappear (without animating). One potential solution is to always render the MessageBanner and pass an enabled prop. Then use `getDerivedStateFromProps` to calculate state on the enabled prop change.

android/build.gradle
```diff
@@ -16,7 +16,7 @@ buildscript {
     }
 
     dependencies {
-        classpath "com.android.tools.build:gradle:${versions.gradleBuildTools}"
+        classpath 'com.android.tools.build:gradle:3.1.3'
```
RGBboy: This needs to be reverted 👍

android/gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,5 +1,6 @@
+#Thu Jul 05 11:56:36 BST 2018
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
 zipStoreBase=GRADLE_USER_HOME
 zipStorePath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-4.2.1-all.zip
+distributionUrl=https\://services.gradle.org/distributions/gradle-4.4-all.zip
```
RGBboy: The changes in this file need to be reverted too.

src/components/MessageBanner.test.js
```diff
@@ -0,0 +1,122 @@
...
+  it("should be able to reset initial animation value to 0", () => {
+    const output = render(defaultProps);
+    const instance = output.instance();
+
+    instance.resetAnimation();
+    expect(timingSpy).toHaveBeenCalledTimes(1);
+    expect(timingSpy.mock.calls[0][1].toValue).toEqual(0);
+  });
+
+  it("should reset isAnimating to false after animation is complete", () => {
```
RGBboy: This test contains a bunch of comments. Can you fix it up just to have the test you want?

src/screens/ParadeMapScreen/Map.js
```diff
@@ -119,10 +119,8 @@ class Map extends PureComponent<Props, State> {
         this.setState({
           moveToUserLocation: false
         });
-        Alert.alert(
-          "We couldn't find your location",
-          "GPS or other location finding magic might not be available, please try again later"
-        );
+
+        this.messageBannerRef.current.handleAnimation();
```
RGBboy: My only suggestion here would be to rename handleAnimation to something more appropriate for external consumers of the MessageBanner class. For example `show()`.

PR #576, opened 2018-07-04
Stage marker layering behaviour

### LINKED ISSUE #567: Active state of stages don't always appear above other pins (iOS)
## Description of issue
When selecting stages that have overlapping amenities, the active stage then appears underneath the surrounding amenities

## Expected behaviour
The active pin should appear at the top level

If it's a functional issue, describe what would have been the correct behaviour.

## Steps to reproduce

List the exact steps you took to arrive at the issue.
Open the map
Select a stage that has overlapping amenities

iOS - iPhone X
Not tested on other iOS devices

## Screenshots
img_9730
img_9731

Lily418:
This PR is to fix #567 and improve behaviour of the layering of the markers. If an amenity is selected the currently selected stage should be unselected and that amenity marker should be drawn above. Otherwise stages appear above amenities and a selected stage should appear above everything else. 

I followed this [Z Index Marker](https://github.com/react-community/react-native-maps/blob/master/example/examples/ZIndexMarkers.js) example to apply the z indexes. 

It was also useful to see the way react native maps manipulates the z indexes based on open callouts.

https://github.com/react-community/react-native-maps/blob/080678b24f886c3b8104206f2f80452ee723243a/lib/ios/AirMaps/AIRMapMarker.m#L316

## Tester check-list

  * [ ] Galaxy S7 (Android 6.0)
  * [ ] iPhone X (iOS 11.x)
  * [ ] iPhone 8 (11.x)
  * [ ] iPhone 7 (iOS 10.x)

  Optional:

  * [ ] Google Pixel 2
  * [ ] Galaxy S8 Plus
  * [ ] Galaxy S7 Edge
  * [ ] Galaxy S6
  * [ ] iPhone 7+
  * [ ] iPhone 6

### DIFF
--- a/src/screens/ParadeMapScreen/Map.js
+++ b/src/screens/ParadeMapScreen/Map.js
@@ -151,7 +151,7 @@ class Map extends PureComponent<Props, State> {
     this.setState({ tileDetails: stage, activeMarker: stage.id });
   };
 
-  handleMapPress = () => {
+  dismissEventTile = () => {
     this.setState({ tileDetails: null, activeMarker: null });
   };
 
@@ -201,7 +201,7 @@ class Map extends PureComponent<Props, State> {
           showsBuildings={false}
           showsTraffic={false}
           showsIndoors={false}
-          onPress={this.handleMapPress}
+          onPress={this.dismissEventTile}
         >
           <Polyline
             coordinates={this.props.route}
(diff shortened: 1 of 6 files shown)


robbiemccorkell:
Amazing thank you! I couldn't figure out how to get those Z indexes to work right.

PR #580, opened 2018-07-05
remove end date from event details screen

### LINKED ISSUE #330: Event details page: Remove the end date when events run past midnight
## Description of issue

The Event details page currently shows a date range when events run past midnight.

## Expected behaviour

Only show the start date as events aren't intended to run over several days and it should be obvious to users that some events will run into the early hours of the following day.

RGBboy:
Fixes #330. Removes end date from event details screen.

screen shot 2018-07-05 at 10 30 30

### DIFF
--- a/src/screens/EventDetailsScreen/EventOverview.js
+++ b/src/screens/EventDetailsScreen/EventOverview.js
@@ -2,7 +2,6 @@
 import React from "react";
 import {
   toLondonFormat as formatDate,
-  isSameDay,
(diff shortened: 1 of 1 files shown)


Lily418:
Should their be support for events which run longer than 2 days? This change makes sense for events which run beyond midnight but could be confusing for multiday events.

Lily418:
Ah yes saw "events aren't intended to run over several days" in this bug report.

PR #583, opened 2018-07-05
Message bar component rxjs fix

RGBboy:
Update the implementation of `locationStatusStream` to timeout after 3000ms of no messages. This also adds tests around this using `rxjs` TestScheduler and its marble diagrams 😱. See this for more info: https://github.com/ReactiveX/rxjs/blob/master/doc/marble-testing.md


RGBboy:
@rgowan This is the PR that should also fix the android TimeoutError that we were getting on Android. Need to double check though.

Issue #587, opened 2018-07-06
Optimize image loading


RGBboy:
Contentful has an API to crop and resize images. Currently we just load the original image regardless of size. To help with the Android out of memory issue and decrease the bandwidth consumption of the app we should load the appropriate sized image for where we are displaying it. As an example [Cabaret_finalists.jpg?w=500&h=500&fit=fill](https://images.ctfassets.net/n2o4hgsv6wcx/4iXietXWjecIke4K4m8mii/f51a9ba4d860859cc3286d23cc29c0e7/Cabaret_finalists.jpg?w=500&h=500&fit=fill).

This image originally returns the following data:

```
{
  id: "4iXietXWjecIke4K4m8mii"
  revision: 1
  uri: "https://images.ctfassets.net/n2o4hgsv6wcx/4iXietXWjecIke4K4m8mii/f51a9ba4d860859cc3286d23cc29c0e7/Cabaret_finalists.jpg"
  width: 1500
  height: 1000
}
```

We would need to figure out the size of image to request. Perhaps by using React Natives onLayout handler (although not using this API would be preferred).

robbiemccorkell:
😱 I had no idea this was available

Lily418:
Apologies, I'd misunderstood that the Sync API doesn't sync the image data just the url of the images so was looking in the wrong place.

PR #593, opened 2018-07-06
Optimize Contentful image loading

Lily418:
Implemention for #587 

Render transparent image initally and then request an image of the rendered size.

## Tester check-list

  * [ ] Galaxy S7 (Android 6.0)
  * [ ] iPhone X (iOS 11.x)
  * [ ] iPhone 8 (11.x)
  * [ ] iPhone 7 (iOS 10.x)

  Optional:

  * [ ] Google Pixel 2
  * [ ] Galaxy S8 Plus
  * [ ] Galaxy S7 Edge
  * [ ] Galaxy S6
  * [ ] iPhone 7+
  * [ ] iPhone 6

### DIFF
--- a/src/components/ImageBackground.js
+++ b/src/components/ImageBackground.js
@@ -21,9 +21,11 @@ export const ImageBackground = ({
   reference,
   getImageDetails,
   ...props
-}: Props) => (
-  <RNImageBackground source={getImageDetails(reference.sys.id)} {...props} />
-);
+}: Props) => {
+  const imgSrc = getImageDetails(reference.sys.id, { width: 50, height: 25 });
+  console.log("ImageBackground source", imgSrc);
+  return <RNImageBackground source={imgSrc} {...props} />;
+};
 
 // Note we must add a return type here for react-redux connect to work
 // with flow correctly. If not provided is silently fails if types do
--- a/src/components/ImageBackground.js
+++ b/src/components/ImageBackground.js
@@ -23,13 +23,13 @@ export class ImageBackground extends Component<Props, State> {
     this.state = { imageSize: null };
   }
 
-  onLayout({
+  onLayout = ({
     nativeEvent: {
       layout: { width, height }
     }
-  }) {
+  }) => {
     this.setState({ imageSize: { width, height } });
-  }
+  };
 
   render() {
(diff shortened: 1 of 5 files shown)


src/data/image.js
```diff
@@ -15,8 +15,21 @@ export type ImageDetails = {
...
-): ?ImageDetails => images[id];
+  id: string,
+  dimensions: ?{ width: number, height: number }
+): ?ImageDetails => {
+  if (!dimensions) {
+    return images[id];
+  }
+
+  const imageDetails = images[id];
+  const imageUriWithResizingBehaviourParameters = `${imageDetails.uri}?w=${
```
frigus02: I would not count on the image details being present all the time, i.e. `imageDetails` might be `undefined`.

I think it might be possible in Contentful to select an image in an event, which is still a draft. This means we get the event with the image id. But we don't load the image because we only download published images.

src/components/ImageBackground.js
```diff
@@ -11,19 +11,52 @@ type OwnProps = {
...
-}: Props) => (
-  <RNImageBackground source={getImageDetails(reference.sys.id)} {...props} />
-);
+export class ImageBackground extends Component<Props, State> {
+  constructor(props: Props) {
+    super(props);
+    this.state = { imageSize: null };
+  }
+
+  onLayout = (event: Object) => {
```
frigus02: React Native exposes the flow type for the layout event. You could use this here.

Example:

https://github.com/redbadger/pride-london-app/blob/1b40ff6b61b40e00eaacc0a87daa59b220a2325d/src/components/Collapsible.js#L7

https://github.com/redbadger/pride-london-app/blob/1b40ff6b61b40e00eaacc0a87daa59b220a2325d/src/components/Collapsible.js#L65-L67

frigus02:
@Lily418 Btw, you might want to [update your git settings for `user.email`](https://help.github.com/articles/setting-your-commit-email-address-in-git/) so GitHub can link your commits to your profile. Currently your author is `Lily Hoskin <lily@Lilys-MacBook-Pro.local>` 🙂

Lily418:
@frigus02 Thank you, have updated with a new test for that case and updated based on other comments

PR #594, opened 2018-07-06
add same background to App as Spash on android;

### LINKED ISSUE #591: Loading screen: White screen appears half way through loading
## Description of issue

When I first open the app I see the purple loading screen the suddenly the screen whites out, then the purple screen returns and transitions into the app as expected. 

This seems to be after updating the app.

## Expected behaviour

I'd expect there to be an uninterrupted flow from loading, animation and transition into the app.

## Steps to reproduce

Update and open app
Observe during the loading sequence a white screen appears half way through
The loading screen resumes and transitions into the app

Steve's Pixel

## Screenshots

None

RGBboy:
Fixes #591. This should fix the white screen flash Android users see when updating the app. This does not seem to happen when opening an already installed app (that has been opened before) **Note:** There is still a fade of solid color. I'm not sure how to get rid of this.

In order to test you will need to uninstall the app an reinstall it.

## Tester check-list

* [ ] Tested on a number of Android devices:

  * [ ] Galaxy S7 (Android 6.0)

  Optional:

  * [ ] Google Pixel 2
  * [ ] Galaxy S8 Plus
  * [ ] Galaxy S7 Edge
  * [ ] Galaxy S6

### DIFF
--- a/android/app/src/main/res/values/styles.xml
+++ b/android/app/src/main/res/values/styles.xml
@@ -3,6 +3,7 @@
     <!-- Base application theme. -->
     <style name="AppTheme" parent="Theme.AppCompat.Light.NoActionBar">
         <!-- Customize your theme here. -->
+        <item name="android:windowBackground">@drawable/background_splash</item>
     </style>
 
 


stevecottle:
Checked this and is an improvement on the white 👍

frigus02:
Hm, I think the better option would be to set the SplashTheme as the application/android:theme here:

https://github.com/redbadger/pride-london-app/blob/0eb68ea5344805ff2b3710b3febb65883b38d612/android/app/src/main/AndroidManifest.xml#L18

Because we use the clean AppTheme once the SplashScreen activity has been loaded:

https://github.com/redbadger/pride-london-app/blob/0eb68ea5344805ff2b3710b3febb65883b38d612/android/app/src/main/java/org/prideinlondon/festival/MainActivity.java#L12

I think this is important in case we have any transparent views in React Native.

